PR #1, opened 2015-11-14
Make header file compatible with C++

zhm:
This patch puts the public exports in an `extern "C"` block so C++ won't mangle the symbols.


lfittl:
Looks good, merged!

PR #2, opened 2015-11-14
Add API for free'ing results

zhm:
This patch adds a public API for `free()`ing the results from the parse and normalize functions. This allows the caller to only need to use 1 function to clean up the results of the functions and also lets the library expand the result structures without creating leaks in the calling program.

This is mostly a suggestion of an API, I wanted to push it to get feedback. This patch also includes #1.

Thanks!

### DIFF
--- a/pg_query.h
+++ b/pg_query.h
@@ -26,6 +26,8 @@ extern "C" {
 void pg_query_init(void);
 PgQueryNormalizeResult pg_query_normalize(char* input);
 PgQueryParseResult pg_query_parse(char* input);
+void pg_query_free_normalize_result(PgQueryNormalizeResult result);
+void pg_query_free_parse_result(PgQueryParseResult result);
 
 #ifdef __cplusplus
 }
--- a/pg_query_normalize.c
+++ b/pg_query_normalize.c
@@ -358,3 +358,14 @@ PgQueryNormalizeResult pg_query_normalize(char* input)
 
 	return result;
 }
+
+void pg_query_free_normalize_result(PgQueryNormalizeResult result)
+{
+  if (result.error) {
+    free(result.error->message);
+    free(result.error->filename);
+    free(result.error);
+  }
+
+  free(result.normalized_query);
+}
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ int main() {
   result = pg_query_parse("SELECT 1");
 
   printf("%s\n", result.parse_tree);
+
+  pg_query_free_parse_result(result);
 }
 ```
 
--- a/examples/normalize.c
+++ b/examples/normalize.c
@@ -11,13 +11,11 @@ int main() {
 
   if (result.error) {
     printf("error: %s at %d\n", result.error->message, result.error->cursorpos);
-    free(result.error->message);
-    free(result.error);
   } else {
     printf("%s\n", result.normalized_query);
   }
 
-  free(result.normalized_query);
+  pg_query_free_normalize_result(result);
 
   return 0;
 }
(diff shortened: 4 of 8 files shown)


pg_query_normalize.c
```diff
@@ -358,3 +358,14 @@ PgQueryNormalizeResult pg_query_normalize(char* input)
...
 }
+
+void pg_query_normalize_free(PgQueryNormalizeResult result)
+{
+  if (result.error) {
+    free(result.error->message);
+    free(result.error->filename);
+  }
+
+  free(result.error);
```
lfittl: Stylistically speaking, I'm wondering if this shouldn't be inside the if clause.

`free()` doesn't do anything on a NULL pointer, but it'd feel more logical to me if its in there.

pg_query.h
```diff
@@ -19,8 +19,18 @@ typedef struct {
...
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 void pg_query_init(void);
 PgQueryNormalizeResult pg_query_normalize(char* input);
 PgQueryParseResult pg_query_parse(char* input);
+void pg_query_normalize_free(PgQueryNormalizeResult result);
+void pg_query_parse_free(PgQueryParseResult result);
```
lfittl: Should we maybe call them `pg_query_free_normalize_result` and `pg_query_free_parse_result` instead?

(but also fine with the suffix variant, no strong opinion here)

lfittl:
@zhm Excellent work & thinking!

This should make the other callers ([Ruby](https://github.com/lfittl/pg_query.go/blob/master/parse.go#L27) and [Go](https://github.com/lfittl/pg_query/blob/master/ext/pg_query/pg_query_ruby.c#L53)) easier to read as well :)

If you want to address the comments (but fine to leave as-is too), and rebase, then I'd be happy to merge!

Thanks for your contribution 🙇

lfittl:
Ah, just saw you're working on https://github.com/zhm/node-pg-query-native - thats awesome!

I've been thinking quite some time how we can port the other methods in the Ruby API (most importantly `deparse`, so you can turn the AST into SQL again) to other languages.

What are your plans / use case for the node version?

Happy to help in any way I can :smiley:

zhm:
I rebased my patch with your feedback. Thanks!

As for my plans, I do have a need for the deparser because I would like to modify SQL and then deparse it back to run it. I started another little node module that sits on top of https://github.com/zhm/node-pg-query-native to handle some additional things like the deparsing. I do think it would be worthwhile to not need to reimplement the deparsing logic in multiple languages. It's time-consuming and it'd be unfortunate to fix the same bug a bunch of times.

It's also unfortunate to have to implement a deparsing in C/C++, but it might be worth it in the long run.

zhm:
An API like:

```c
typedef struct {
  char* query;
  char* error;
} PgQueryDeparseResult;

PgQueryDeparseResult pg_query_deparse(char* json);
void pg_query_free_deparse_result(PgQueryDeparseResult result);
```

This API would require a JSON parser, but maybe that's acceptable.

lfittl:
Looks good, thanks for the contribution!

lfittl:
@zhm Regarding deparsing:

My current thinking is that it might be better to generate the deparser for every language based on a shared DSL - its really just a series of if statements, loops and string appending.

The upside of doing this inside Postgres (i.e. going back to the C structs) might be that we can re-use the deparsing logic that started being added in 9.5/9.6 - but from my recent looking into this its not as useful as hoped.

I'll see that I take the generative DSL-based approach a try today/tomorrow - I've been meaning to dig into this for a while, and you just gave me a good reason to do so again :smile: - I'll keep you in the loop.

zhm:
That would be awesome!

PR #6, opened 2015-12-16
Updated inputs as const char* for correctness

carymrobbins:
Trivial change, but this makes it nicer when passing in `const char*` as input since these functions don't change the pointer.


lfittl:
@carymrobbins Great minds think alike - the latest `master` should actually already have this in place!

lfittl:
Closing this as in my understanding we have this in master. Please re-open if I misunderstood!

PR #11, opened 2016-05-17
Escape JSON strings so all characters can be represented.

zhm:
This PR fixes a bug where the returned JSON string doesn't correctly encode control characters. It results in certain inputs being unparseable by regular JSON parsers. The simplest way to reproduce is probably just open up `irb` with the ruby gem:

``` ruby
PgQuery.parse "SELECT parse_ident(E'\"c\".X XXXX\002XXXXXX')"
```

The above SQL is from the 9.5.3 regression tests.

This new function was taken directly from the postgres source [here](https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/json.c#L2428). It's basically identical to the `_outToken` we already have except it includes the case for escaping control characters. I figured it'd be best to just use exactly what's in postgres. I first attempted to actually use that exported API except it quickly turned into a rabbit hole of things to drag in because json.c has some server-y stuff in it too. It's pretty unlikely this function to change much so to keep it simple I just copied the function body into the project.

The PR should include a test but I wasn't sure where the tests are being maintained - they look like they're being auto-generated.

### DIFF
--- a/pg_query_json.c
+++ b/pg_query_json.c
@@ -107,32 +107,52 @@ removeTrailingDelimiter(StringInfo str)
 }
 
 static void
-_outToken(StringInfo str, const char *s)
+_outToken(StringInfo buf, const char *str)
 {
-	if (s == NULL)
+	if (str == NULL)
 	{
-		appendStringInfoString(str, "null");
+		appendStringInfoString(buf, "null");
 		return;
 	}
 
-	appendStringInfoChar(str, '"');
-	while (*s)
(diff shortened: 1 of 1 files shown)


lfittl:
@zhm Good catch - we probably should rely on `json.c` more, but thats a bigger project as you realized.

Right now the parse tests are written manually, but the fingerprint tests are generated from a JSON file shared with the other libraries. Will probably consolidate this at some point. I'll add a quick test after merging.

PR #12, opened 2016-05-23
Use LLVM to extract parser from PostgreSQL server source, add PL/pgSQL parsing

lfittl:
## TODO
- [x] Fix the other scripts/ helpers to use the new Postgres source location
- [x] Consider copying only the include files that are actually needed
- [x] Remove (some) comments to reduce download size


zhm:
Wow! :tada:

lfittl:
@zhm :)

Let me know if there's any issues the new approach causes for you - it shouldn't affect you too much, since you've been using pre-built libraries already, if I remember correctly.

For the PL/pgSQL parsing, in case thats interesting, feedback very welcome!

PR #16, opened 2016-07-23
build js module with emscripten

nothingisdead:
This adds a the ability to build this project using emscripten. I did have to modify pg_config.h to disable a few options in order to get it to compile, but I don't think they should affect the parsing at all.  I'm not really very familiar with C/C++, so please let me know if there is a better way to do some of these things and I'll update this pull request.

Thanks for your consideration!

### DIFF
--- a/EMSCRIPTEN.md
+++ b/EMSCRIPTEN.md
@@ -1,0 +1,6 @@
+# Emscripten fork of [libpg_query](https://github.com/lfittl/libpg_query)
+
+## Build Instructions:
+
+- `$ emmake make build`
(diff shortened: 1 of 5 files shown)


lfittl:
@nothingisdead Thanks for your contribution!

I was actually exploring the same at https://github.com/lfittl/pg_query_emscripten - my fault for not referencing that in this project's README.

Ideally we could merge the two approaches - I like your approach to bindings as opposed to calling into C directly.

Do you want to give it a try if you can apply this PR over in that repository?

(to get started: `make update_source` fetches this library's source, fixes `pg_config.h` and calls emscripten)

nothingisdead:
@lfittl I can't believe I didn't see that! It's crazy how similar our approaches ended up being. Yours definitely looks cleaner though. I'll try to get just the C++ bindings working in that repo.

Thanks!

PR #17, opened 2016-09-18
Update README.md

gonzazoid:
I had already written my own wrapper for Node before I found that there is similar project. If you add this information it would be great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ C library for accessing the PostgreSQL parser outside of the server.
 
 This library uses the actual PostgreSQL server source to parse SQL queries and return the internal PostgreSQL parse tree.
 
-Note that this is mostly intended as a base library for https://github.com/lfittl/pg_query (Ruby) and https://github.com/lfittl/pg_query.go (Go).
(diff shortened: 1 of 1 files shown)


lfittl:
@gonzazoid Good catch, and sorry this wasn't more obvious. Merged!

I'm curious - what are you building using the Postgres parser in Node?

gonzazoid:
> Good catch, and sorry this wasn't more obvious. Merged!

Never mind! Thank you ever so much!

> I'm curious - what are you building using the Postgres parser in Node?

It will be (I hope) a type generator for Typescript. It will generate types based on sql function's signatures and tables definitions and generate whole client to access to db (similar code for client side to access to API based on swagger's spec's already written and work). Also I hope to write diff-like module to compare sql expressions based on AST, it would be more useful than bare diff with line-based comparing (as part of migration toolchain).

lfittl:
@gonzazoid Neat, that sounds like an impressive project, if you can make it work :-)

Let me know how it goes, happy to help with any questions you have regarding the parse tree format, or any other things you run into!

gonzazoid:
> if you can make it work :-)

Hope, hope, hope!!! :)

Thank you for your kind words and support!

Issue #26, opened 2017-05-11
Build fails on 32 bit system


herwinw:
```
qnet@devel-o9:~/tmp/libpg_query$ CC=/usr/bin/gcc-4.9 make
compiling src/pg_query.c
In file included from ./src/postgres/include/c.h:53:0,
                 from ./src/postgres/include/postgres.h:47,
                 from src/pg_query_internal.h:4,
                 from src/pg_query.c:2:
./src/postgres/include/pg_config.h:725:24: error: ‘__int128’ is not supported for this target
 #define PG_INT128_TYPE __int128
                        ^
./src/postgres/include/c.h:298:9: note: in expansion of macro ‘PG_INT128_TYPE’
 typedef PG_INT128_TYPE int128;
         ^
In file included from ./src/postgres/include/postgres.h:47:0,
                 from src/pg_query_internal.h:4,
                 from src/pg_query.c:2:
./src/postgres/include/c.h:298:24: warning: type defaults to ‘int’ in declaration of ‘int128’ [-Wimplicit-int]
 typedef PG_INT128_TYPE int128;
                        ^
In file included from ./src/postgres/include/c.h:53:0,
                 from ./src/postgres/include/postgres.h:47,
                 from src/pg_query_internal.h:4,
                 from src/pg_query.c:2:
./src/postgres/include/pg_config.h:725:24: error: ‘__int128’ is not supported for this target
 #define PG_INT128_TYPE __int128
                        ^
./src/postgres/include/c.h:299:18: note: in expansion of macro ‘PG_INT128_TYPE’
 typedef unsigned PG_INT128_TYPE uint128;
                  ^
Makefile:73: recipe for target 'src/pg_query.o' failed
make: *** [src/pg_query.o] Error 1
```

lfittl:
Hi @herwinw!

The library code is currently pre-configured for 64-bit systems only, mostly to avoid having a dependency on autoconf for the typical users.

Whats your use case - are you looking to use this library directly, or one of the helper libraries in a scripting language?

herwinw:
I got a review request for a change in a SQL file, which contained both indentation fixes and real changes, and I wanted to simplify the review. Due to legacy reasons (and I surely don't mean lazyness...) we're still at 32 bit.

[...]

PR #27, opened 2017-05-15
Removed references to 128 bit

herwinw:
This makes it compatible with 32 bit systems. We don't need use support
for 128 bit integers anywhere in the codebase.

An alternative option would be to make this depend on a configure script
(like postgres does to generate pg_config.h), but this seems overkill
given we don't really have need for them.

This fails at my local machine when running the concurrency tests:
```
ERROR creating pthread - pthread_create return code 11
```
which doesn't appear to be related to the change itself. This should fix the problem in #26.

### DIFF
--- a/src/postgres/include/pg_config.h
+++ b/src/postgres/include/pg_config.h
@@ -721,9 +721,6 @@
 /* Define to the version of this package. */
 #define PACKAGE_VERSION "9.5.3"
 
-/* Define to the name of a signed 128-bit integer type. */
-#define PG_INT128_TYPE __int128
-
 /* Define to the name of a signed 64-bit integer type. */
 #define PG_INT64_TYPE long int
 


src/postgres/include/c.h
```diff
@@ -289,17 +289,6 @@ typedef unsigned long long int uint64;
...
- *		There currently is only a limited support for the type. E.g. 128bit
- *		literals and snprintf are not supported; but math is.
- */
-#if defined(PG_INT128_TYPE)
-#define HAVE_INT128
-typedef PG_INT128_TYPE int128;
-typedef unsigned PG_INT128_TYPE uint128;
-#endif
-
-/*
```
lfittl: Do we actually need to remove this, given PG_INT128_TYPE is not defined now? (would be good if you can test, since its easier if the change is just in pg_config.h due to code generation)

herwinw: No, we don't have to, we're just making the life of the preprocessor a bit easier. Do you want me to put it back?

(I'm reading "Walden" at the moment, so I might be a bit too much in a mindset of stripping everything non-essential :) )

lfittl: Heh, the only reason its useful is because we can then just teach the automatic code generation about this :)

Specifically, we should be able to take care of the whole deal by adding:

```echo "#undef PG_INT128_TYPE" >> ./src/postgres/include/pg_config.h```

after this line: https://github.com/lfittl/libpg_query/blob/47e6956c0d1182fcce4a23fd04ee3626887c9f10/Makefile#L70

Happy to take care of adding it there post-merge though, since you need a few more tools to run "extract_source" :)

herwinw: It has been updated, the only changes are in `pg_config.h` now

PR #28, opened 2017-05-21
Move inclusion of stdio.h in plpgsql parser

herwinw:
We have to define _GNU_SOURCE before we include this file to get access
to asprintf. In case any other inclusion before already included stdio.h
the define line is effectively ignored, which may result in warnings
about calling the undefined method asprintf.

It would have been enough to move the define line to the top, but by
keeping the define and include as close to eachother as possible it's
clearer why it's there.


lfittl:
Nice, good catch!

PR #31, opened 2017-05-21
Improve concurrency tests

herwinw:
* Use `perror` to display the error in case the thread could not be created
* Lower the number of threads, 1000 made the test fail on my system with a lack of resources, which doesn't tell us anything about the actual code.


lfittl:
Nice, thanks for the contribution!

PR #35, opened 2017-08-09
Rectify the output of the sample program

lelit:
The `README.md` file shows what I think is an outdated output of the sample C program.


lfittl:
Yes, you are indeed correct - thanks for noticing and fixing!

Issue #40, opened 2017-12-17
Problems with linking of headers/libraries?


D93:
Hello Lukas!
I installed libpg_query just as described in Readme/Wiki.
Now I'm facing a problem with compiling the examples because of undefined reference to some functions, regardless of strictly following your instructions about compiling. 

dmitry@dmitry-Aspire-V3-571G ~/libpg_query/examples $ cc -I../ -L../ -lpg_query simple.c

/tmp/ccx5VZ4i.o: In Funktion `main':
simple.c:(.text+0x29): Nicht definierter Verweis auf `pg_query_parse'
simple.c:(.text+0x7f): Nicht definierter Verweis auf `pg_query_free_parse_result'
collect2: error: ld returned 1 exit status

Why do I get this linking error? 
Thanks you very much in advance for your answer.

lelit:
CC arguments order matter, in this particular case the _link_ option referring to the library should go after the source containing references to symbols defined there.

Consider:

```
$ cc -I.. -L.. -lpg_query simple.c
/tmp/cc8zG7EM.o: In function `main':
simple.c:(.text+0x34): undefined reference to `pg_query_parse'
simple.c:(.text+0x7f): undefined reference to `pg_query_free_parse_result'
collect2: error: ld returned 1 exit status
```

and

```
$ cc -I.. -L.. simple.c -lpg_query
$ ls -l a.out 
-rwxrwxr-x 1 lele lele 2738688 Dec 17 17:30 a.out
```

D93:
@lelit:
Thank you very much for your answer as it solves this issue!

@lfittl:
You may want to update the comment headers in the c-files of your examples.
With minimal changes it could be something like this (changed argument order of cc):

// Welcome to the easiest way to parse an SQL query :-)
// Compile the file like this:
//
// cc -I../ -L../ simple.c -lpg_query

lfittl:
@D93 Done, per the PR that @lelit sent in - thanks for the report!

PR #41, opened 2017-12-17
Fix compilation example command, moving the library after the source …

lelit:
…unit

See issue #40.


lfittl:
@lelit Thanks for the fix!

PR #55, opened 2019-04-01
DuckDB uses a simplified version of libpg_query

hannes:


lfittl:
@hannesmuehleisen Sorry for the late response here, happy to see you're using it!

When you say "simplified" - is there a way we could keep a common branch, to make your work to rebase on new Postgres versions easier?

hannes:
We made a lot of changes, including Windows support and exception handling **without** `setjmp` and friends, memory management, error handling etc. Also greatly simplified the header zoo that PG produced. I am not sure how easy it is to update this with newer `libpg_query` releases. Code is here: https://github.com/cwida/duckdb/tree/master/third_party/libpg_query

lfittl:
@hannesmuehleisen This looks great!

I think you are correct that this might be difficult to unify. However, this is a great example of why the parser needs to be accessible outside of Postgres.

Did you make any modifications to the parser logic itself (e.g. to support custom syntax), or is the grammar/keywords/etc still the same?

hannes:
Grammar and keywords are unchanged

PR #57, opened 2019-06-08
Ignore return value of asprintf without compiler warnings

herwinw:
```
src/pg_query_parse_plpgsql.c: In function ‘pg_query_parse_plpgsql’:
src/pg_query_parse_plpgsql.c:433:8: warning: variable ‘ignored’ set but not used [-Wunused-but-set-variable]
    int ignored;
        ^~~~~~~
```


lfittl:
Excellent, thanks for fixing this properly!

PR #58, opened 2019-07-26
More flexible makefile, richer nodes, source update

ethanresnick:
Apologies for bundling a few unrelated changes; if needed, I can split this PR up. But I think they all may be uncontroversial, so it was easier first to just submit them together.

Commit by commit summary:

- a1800b0: Allow the user to provide extra compiler flags when compiling the source. These will get concatenated with the flags already in the Makefile. I needed this, but I imagine it could be generally useful. By using `override`, it does mean that users can't _remove_ flags specified in the makefile, because any flags the user specifies will get merged in rather than replacing the existing ones. But I imagine that's ok, and more convenient overall than requiring the user to copy all the compile argument that are in the makefile just to add new ones.

- 9df7989 and 9a36228: Expose a couple more fields on some PLPgSQL nodes.

- 2ed9034: It looks like the custom `compile_create_function_stmt` function was based off of the [`plpgsql_compile_inline`](https://github.com/postgres/postgres/blob/REL_10_STABLE/src/pl/plpgsql/src/pl_comp.c#L794) function. However, that function in the upstream has changed a bit since `compile_create_function_stmt` was written, so this is just me backporting some of those changes to keep the sources more in sync. I don't think there's any functional difference. Note that I'm backporting from the code as it looks in the v10 branch upstream, to match this library. (It's changed more in v11/v12.)


lfittl:
@ethanresnick Thanks for this PR - I've incorporated your work here into the Postgres 13 release, and attributed you in the commit messages as the author. Appreciate the contribution!

PR #62, opened 2019-11-22
Update to 12-latest

rdunklau:
Hello,

I have taken the time to port libpg_query to PG12.

To do that reliably, I overhauled the whole building process. Instead of parsing the PostgreSQL source code using clang and generating a stripped down version, we trust the compiler to do the right thing.

So, this implies: 
  - patching some definitions to set the weak attribute, allowing them to be overriden in a mock file
  - patching the postgres Makefile to generate a big archive file with everything we need
  - patching some selected global variables to make them thread-safe.
  - compiling with the correct options to allow for the maximum amount of dead code elimination

I understand this is a major change for the project, but this should prove to be more maintainable since we wouldn't have to worry about clang AST format change, and reduces the repo size tremendously.


PierreDucroquet:
FYI, this use heavily the LTO done by the compiler and the linker to strip down the binary.
An alternative path, but much harder to implement, would have been to compile to LLVM bitcode instead of regular object files, but this is much less portable.

lelit:
@lfittl, did you have a chance to take a look to this PR, to make a pronouncement?
From my POV it seems a good step toward making future maintenance easier...

purcell:
This looks like a nice piece of work — it looks like @lfittl has been too busy to work on this library since October, and the last release was in May 2018, so perhaps he would appreciate help maintaining it? FWIW, I've personally had good success with routinely granting commit access to contributors who submit high-quality changes like this PR: I learned this trick from some very effective open source maintainers.

mikebaldry:
I agree this is a better way forward and it would be awesome if @lfittl would be able to take a look so we can more forward.

I do have a problem with this method though

I am unable to get this to build on macOS Catalina. There seems to be a variety of issues, most of which I am not equip to resolve :) 

I can confirm running `./configure && make` on the unpatched postgres source works correctly

but running this make, I hit a variety of problems:

- replaced `LDFLAGS` `-Wl,--gc-sections,--as-needed` to `-Wl,-dead_strip` (suggestion on stackoverflow). `./configure` failed with `ld: unknown option: --gc-sections`

- removed `AR="$(AR)"` from `./configure` line, as I get `make[1]: gcc-ar: No such file or directory`. I'm not sure if there is a way to get `configure` to decide what this should be?

Now I am stuck at this point:
```
gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -Wno-unused-command-line-argument -fPIC -flto -fvisibility=hidden -fdata-sections -ffunction-sections -Os -pthread -Os -g  -bundle -multiply_defined suppress -o pgcrypto.so pgcrypto.o px.o px-hmac.o px-crypt.o crypt-gensalt.o crypt-blowfish.o crypt-des.o crypt-md5.o md5.o sha1.o internal.o internal-sha2.o blf.o rijndael.o pgp-mpi-internal.o imath.o mbuf.o pgp.o pgp-armor.o pgp-cfb.o pgp-compress.o pgp-decrypt.o pgp-encrypt.o pgp-info.o pgp-mpi.o pgp-pubdec.o pgp-pubenc.o pgp-pubkey.o pgp-s2k.o pgp-pgsql.o  -L../../src/port -L../../src/common  -Wl,-dead_strip  -Wl,-dead_strip_dylibs   -bundle_loader ../../src/backend/postgres 
ld: file too small (length=0) file '../../src/backend/postgres' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [pgcrypto.so] Error 1
make: *** [/Users/mikeyb/projects/libpg_query/tmp/postgres] Error 2
```

I can't work out what is causing this or why, and have limited experience with configure/gcc/make

lelit:
Porting this to `clang` will require some knowledge and love: from the quick research I did months ago `clang` seems to implement everything is needed, but unfortunately with a slightly different syntax and tooling.

mikebaldry:
> Porting this to `clang` will require some knowledge and love: from the quick research I did months ago `clang` seems to implement everything is needed, but unfortunately with a slightly different syntax and tooling.

Oh I see. I'm happy to build with whatever is needed, I have given it a go with a version of GCC, but it seems that gnu `ld` isn't ported/compatible with mac, so I get stuck at the same stage still, and there is no other binary I can make it use instead for `ld`.

pyramation:
this is amazing! thank you @rdunklau :) I'm also unable to build on mac currently

pyramation:
while you're at it, if possible to make a request, can we expose the deparser too? Not sure if that's hard but it would be super helpful

lelit:
> can we expose the deparser too?

Is the _deparser_ something on the line of https://github.com/lelit/pglast?

pyramation:
@lelit yes, exactly. Do you also agree it be great to just use the C version? Then it would be portable to node as well as python (which is what pglast looks like).  I spent a lot of time maintaining https://github.com/pyramation/pgsql-parser which is in JS and relies on libpg_query... however, if we exposed the C deparser, we could just write a Node addon and get the real PG doing the work :)

lfittl:
Hi everyone on this PR!

First of all, I'd like to thank everyone for their patience on this (I know its been a long time coming), and thank you @rdunklau for your efforts on making Postgres 12 work with `libpg_query`, many months ago.

I've reviewed the approach taken in this PR carefully, and the reason I haven't merged this is because its a very different approach than how the library is supposed to work. It also has compatibility issues for non-GCC based systems, as discussed earlier.

Benefits of the approach in this PR:
- Resulting object size is slightly smaller (e.g. the "test/parse" binary is 2.3MB instead of 2.8MB)
- No reliance on special LLVM AST handling logic during Postgres release update time

Downsides to the approach in this PR:
- The result is not self-contained (builds utilizing the libpg_query.a file depend on the Postgres source tree being present)
- Long build time (130 seconds vs 20 seconds)
- Does an interactive download as part of the build process (this is what libpg_query used to do, and it caused various issues)
- Requires a new-ish Linux distribution (failed for me on RHEL 6-like Amazon Linux, works on RHEL 7 / Amazon Linux 2)
- Depends on "--gc-sections,--as-needed" identifying unneeded code portions at build time (making the behavior different depending on compiler version, and requiring GCC instead of supporting other compilers like clang)
- "make examples" fails (for simple_plpgsql.c example - with linker errors)
- Parser patch does not seem to work correctly (test/complex fails)

Therefore, I won't be moving forward with the approach taken in this PR for the library.

---

However, I've finally had sufficient time to focus on updating this library to Postgres 12, in the original approach this library has taken. This has definitely been made easier by @rdunklau's work here.

The current work in progress can be found here: https://github.com/lfittl/libpg_query/pull/67

There is still work to be done (in particular, the fingerprinting will see some changes), but the basic functionality works, and I would encourage folks to test that branch out, and give feedback in that PR.

I will leave this PR here open until the other PR is completed and ready for production use.

rdunklau:
Thank you for this feedback.
I will try adress some of the complaints you have.

> - The result is not self-contained (builds utilizing the libpg_query.a file depend on the Postgres source tree being present)
> - Long build time
> - Interactive download

Fair enough, but it seemed an acceptable compromise to have this. The final library being built doesn't have this kind of restrictions. As for the interactive download, that could be solved by shipping the PostgreSQL source tree too (it's MIT, so no problem to fork it wherever needed).

> - Requires a new-ish Linux distribution (failed for me on RHEL 6-like Amazon Linux, works on RHEL 7 / Amazon Linux 2)

Maybe the default GCC version on RHEL6 is too old, but I guess it could be updated. In any case, this is a 10 year old linux distribution which is coming to end of life soon. RHEL7 is already 5 years old and is supported as you noted.

> - Depends on "--gc-sections,--as-needed" identifying unneeded code portions at build time (making the behavior different depending on compiler version, and requiring GCC instead of supporting other compilers like clang)

As far as I know, clang could be supported too since it offers the same kind of LTO features. It just needs some work which we haven't done yet.
> -"make examples" fails (for simple_plpgsql.c example - with linker errors)
> - Parser patch does not seem to work correctly (test/complex fails)

Do you have more information about this ? It works for us, so I'm a bit surprised.

As for the general approach, I think it is far easier to maitain using this approach. As an example, @pinaraf  just pushed a PG13 compatible branch on our repo, which was very quick to do (about 20minutes, I think)

The approach taken with the ruby scripts extracting the relevant portion of the code depends heavily on the LLVM version, whose AST can change at any point.

lfittl:
> > * The result is not self-contained (builds utilizing the libpg_query.a file depend on the Postgres source tree being present)
> > * Long build time
> > * Interactive download
> 
> Fair enough, but it seemed an acceptable compromise to have this. The final library being built doesn't have this kind of restrictions. As for the interactive download, that could be solved by shipping the PostgreSQL source tree too (it's MIT, so no problem to fork it wherever needed).

Let me see if I can give more background on this, to explain this more.

Most of the time, libpg_query is used by bindings in other languages (such as Ruby, Go, Node.js, etc), instead of being used directly.

For these bindings, its important that the default installation mechanism (e.g. "gem install pg_query" in Ruby) works without issues. There is two approaches we can take to support this:

1) Provide an already compiled binary version for popular architectures (e.g. amd64 on Linux + macOS)

2) Provide a source only version (embedding the full Postgres source code, as you mention)

The big problem with (1) is that it doesn't work with all languages. e.g. in Go, what `cgo` allows you to do is very limited, and in my experience having a source tree present is much easier than fiddling with an `.a` file, which would commonly require a Makefile (which doesn't work well with the Go modules system).

The downside to (2) is the higher build times as noted.

> > * Requires a new-ish Linux distribution (failed for me on RHEL 6-like Amazon Linux, works on RHEL 7 / Amazon Linux 2)
> 
> Maybe the default GCC version on RHEL6 is too old, but I guess it could be updated. In any case, this is a 10 year old linux distribution which is coming to end of life soon. RHEL7 is already 5 years old and is supported as you noted.

For sure, and I'll be happy when RHEL 6 is gone 😄  - I don't think this alone would be a blocking issue.

> 
> > -"make examples" fails (for simple_plpgsql.c example - with linker errors)
> > 
> > * Parser patch does not seem to work correctly (test/complex fails)
> 
> Do you have more information about this ? It works for us, so I'm a bit surprised.

It failed on my quick attempt - I'm sure we could resolve it, so this wouldn't be a blocking issue on its own.

> As for the general approach, I think it is far easier to maitain using this approach. As an example, @pinaraf just pushed a PG13 compatible branch on our repo, which was very quick to do (about 20minutes, I think)

FWIW, with a similar amount of effort:

https://github.com/lfittl/libpg_query/pull/68/files

More importantly, these files are also updated:
- `srcdata/*`
- Update fingerprint definitions (note that fingerprinting works based on ignoring certain fields, which means that the code needs to be updated with each release to output everything thats not ignored: https://github.com/lfittl/libpg_query/pull/68/files#diff-19c65a5b9e81820263787173ab712846 -- this change was auto-generated based on the LLVM-parsed source code)
- Add new struct fields to JSON output helpers: https://github.com/lfittl/libpg_query/pull/68/files#diff-4686c411fbeb762f9e725085899f6cb2

In your approach these would need to be manually updated, something which I intentionally tried to avoid by taking the LLVM-based approach.

> The approach taken with the ruby scripts extracting the relevant portion of the code depends heavily on the LLVM version, whose AST can change at any point.

That is correct, and certainly a risk. That said, at least in rebasing for 12 (or 13), the only change I had to do was related to the macOS build environment, not based on any AST changes in LLVM.

---

Overall, one additional item I'm taking away: The LLVM-based approach is not accessible enough, and not documented well at this point. I think this can be improved on (by better documentation, etc), and is something I'll try to spend time on.

Btw, here is a similar idea of using LLVM for `outfuncs` generation in Postgres - Perl-based, but similar idea to how libpg_query does it: https://www.postgresql.org/message-id/flat/20190828234136.fk2ndqtld3onfrrp%40alap3.anarazel.de

rdunklau:
Thank you for spending the time to explain this.
I agree with you that it would be very helpful to have some documentation on the process, especially the different dependencies versions needed to make it work.

lfittl:
Closing this as Postgres 13 support has been released. Thanks!

purcell:
Amazing, thanks @lfittl!

purcell:
I see that the 2.0.0 contains C files which reference postgres.h, but that that file itself is not included in pg_query, and I'm finding that on a system without postgresql installed, the native extension fails to build because of that missing file. I feel like it should still build in this scenario?

lfittl:
@purcell If you're referring to the Ruby gem, that should be fixed in 2.0.1 now (see https://github.com/pganalyze/pg_query/issues/191)

purcell:
> @purcell If you're referring to the Ruby gem, that should be fixed in 2.0.1 now (see [pganalyze/pg_query#191](https://github.com/pganalyze/pg_query/issues/191))

Sorry, yes I was. Thanks!

PR #65, opened 2020-04-23
Fix parallel build

marco44:
Small fix to have the correct build order when using -j option


marco44:
Woops wrong repository sorry

PR #75, opened 2020-11-04
Add libpg_query-sys rust library

FGRibreau:


lfittl:
@FGRibreau Thanks for contributing this!

I've opened https://github.com/tdbgamer/libpg_query-sys/issues/1 to see if the library author is open to changing their library to a BSD-type license. The Mozilla Public License has a few restrictions that make it complicated to use sometimes, and I'd like to make it straightforward for promoted libraries to be used by using BSD/MIT-type licenses.

PR #80, opened 2020-12-29
Add arch-ppc.h for ppc architectures

pkubaj:


lfittl:
Thanks for the contribution!

Merged via https://github.com/lfittl/libpg_query/commit/820f5294f92927d6c4f61722bb28cf380942b83c - note that I regenerated the file to use the correct version for Postgres 10 (this PR has a slightly newer version of the file, which is not needed on older Postgres versions)

This will also be included for the upcoming parser update to Postgres 13.

PR #91, opened 2021-03-20
Table like clause issue

lelit:
Hi Lukas,

while working on my `pglast` _deparser_ I noticed that the following statement
```sql
CREATE TABLE like_constraint_rename_cache   (LIKE constraint_rename_cache INCLUDING ALL)
```

gets deparsed by current `libpg_query` as
```sql
CREATE TABLE like_constraint_rename_cache (LIKE constraint_rename_cache INCLUDING ALL INCLUDING COMMENTS INCLUDING CONSTRAINTS INCLUDING DEFAULTS INCLUDING IDENTITY INCLUDING GENERATED INCLUDING INDEXES INCLUDING STATISTICS INCLUDING STORAGE)
```

It seems redundant, isn't it? If so, it this PR reasonable?

Unfortunately I could not run the tests locally, because a `make` stopped with
```
...
compiling src/postgres/src_port_strerror.c
compiling src/postgres/src_port_strnlen.c
compiling vendor/protobuf-c/protobuf-c.c
compiling vendor/xxhash/xxhash.c
Warning: protoc-gen-c not found, skipping protocol buffer regeneration
compiling protobuf/pg_query.pb-c.c
ar: Cannot convert existing thin library libpg_query.a to normal format
make: *** [Makefile:145: libpg_query.a] Errore 1
```

and I could not figure out what's wrong here... I'm on a Debian sid, if that matters.

### DIFF
--- a/src/pg_query_deparse.c
+++ b/src/pg_query_deparse.c
@@ -4018,23 +4018,25 @@ static void deparseTableLikeClause(StringInfo str, TableLikeClause *table_like_c
 
 	if (table_like_clause->options == CREATE_TABLE_LIKE_ALL)
 		appendStringInfoString(str, "INCLUDING ALL ");
-	if (table_like_clause->options & CREATE_TABLE_LIKE_COMMENTS)
+        else
+        {
+	    if (table_like_clause->options & CREATE_TABLE_LIKE_COMMENTS)
 		appendStringInfoString(str, "INCLUDING COMMENTS ");
-	if (table_like_clause->options & CREATE_TABLE_LIKE_CONSTRAINTS)
+	    if (table_like_clause->options & CREATE_TABLE_LIKE_CONSTRAINTS)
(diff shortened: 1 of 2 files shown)


lfittl:
@lelit Nice, good catch!

I wonder if we should add this as an explicit test to the list in [test/deparse_tests.c](https://github.com/pganalyze/libpg_query/blob/13-latest/test/deparse_tests.c)? (those are checked for matching 1:1 in the text - the regression tests are only checked for matching in the parse tree, but not in the exact text representation)

(I'll review the issue you've reported on running the tests separately - need to boot a VM with Debian sid to confirm)

lelit:
Thanks!

lfittl:
@lelit I've just ran `make` successfully (including the tests) on Debian sid, including running all the tests.

Could it be that you have a leftover `libpg_query.a` library from an older version of libpg_query?

PR #100, opened 2021-04-06
Build also as a shared library and add install target

jirutka:
BTW: I’ve packaged [libpg_query](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/libpg_query) and [ruby-pg_query](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/ruby-pg_query) for Alpine Linux. It’s built with patches from this PR and few others to unbundle dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,11 @@ PROTOC_VERSION = 3.14.0
 
 VERSION = 2.0.4
 VERSION_MAJOR = $(call word-dot,$(VERSION),1)
+VERSION_MINOR = $(call word-dot,$(VERSION),2)
+VERSION_PATCH = $(call word-dot,$(VERSION),3)
 
 SONAME = $(SOLIB).$(shell printf '%02d%02d' $(PG_VERSION_MAJOR) $(VERSION_MAJOR))
+SOLIBVER = $(SONAME).$(VERSION_MINOR).$(VERSION_PATCH)
 
 SRC_FILES := $(wildcard src/*.c src/postgres/*.c) vendor/protobuf-c/protobuf-c.c vendor/xxhash/xxhash.c protobuf/pg_query.pb-c.c
 NOT_OBJ_FILES := src/pg_query_enum_defs.o src/pg_query_fingerprint_defs.o src/pg_query_fingerprint_conds.o src/pg_query_outfuncs_defs.o src/pg_query_outfuncs_conds.o src/pg_query_readfuncs_defs.o src/pg_query_readfuncs_conds.o src/postgres/guc-file.o src/postgres/scan.o src/pg_query_json_helper.o
@@ -50,6 +53,8 @@ CLEANOBJS = $(OBJ_FILES)
 CLEANFILES = $(PGDIRBZ2)
 
 AR = ar rs
+INSTALL = install
+LN_S = ln -s
 RM = rm -f
 ECHO = echo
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -251,3 +267,17 @@ test/scan: test/scan.c test/scan_tests.c $(ARLIB)
...
 	$(CC) $(TEST_CFLAGS) -o $@ test/split.c $(ARLIB) $(TEST_LDFLAGS)
+
+prefix = /usr/local
+libdir = $(prefix)/lib
+includedir = $(prefix)/include
+
+install: $(ARLIB) $(SOLIB)
+	$(INSTALL) -d "$(DESTDIR)"$(libdir)
+	$(INSTALL) -m 644 $(ARLIB) "$(DESTDIR)"$(libdir)/$(ARLIB)
+	$(INSTALL) -m 755 $(SOLIB) "$(DESTDIR)"$(libdir)/$(SOLIBVER)
```
lfittl, line 280: Does the shared library actually need to be executable? (I would have thought 644 is sufficient here)

Makefile
```diff
@@ -1,13 +1,24 @@
...
 PG_VERSION = 13.2
+PG_VERSION_MAJOR = $(call word-dot,$(PG_VERSION),1)
 PROTOC_VERSION = 3.14.0
 
+VERSION = 2.0.4
+VERSION_MAJOR = $(call word-dot,$(VERSION),1)
+VERSION_MINOR = $(call word-dot,$(VERSION),2)
+VERSION_PATCH = $(call word-dot,$(VERSION),3)
+
+SONAME = $(SOLIB).$(shell printf '%02d%02d' $(PG_VERSION_MAJOR) $(VERSION_MAJOR))
```
lfittl: I think we should include the minor version into the SONAME, since we may have ABI breaks with minor releases as well (but it seems reasonable to avoid them with patch releases).

Makefile
```diff
@@ -74,14 +87,14 @@ endif
 
 all: examples test build
 
-build: $(ARLIB)
+build: $(ARLIB) $(SOLIB)
```
lfittl: I'd like to avoid building the shared library by default, since its a special case - the default expectation is that pg_query is used as a static library, and examples/etc are written with that in mind.

I'd recommend we make a separate `build_shared` (or similar) target, and install depends on that as well as `build`.

lfittl (commented):
@jirutka Thanks for the contribution!

Whilst I'm not sure if its a good idea to use libpg_query as a shared library, and it was written with a static library in mind, it doesn't hurt either I think.

My only requests here are to avoid building the shared library by default, and to include the minor version in the SONAME (since it's likely we'll break the ABI in a minor release sometimes).

Makefile
jirutka: Yes, shared libraries are always executable, just look into your `/usr/lib` (or `/usr/lib64`)…

Makefile
jirutka: Resolved by ade04b03f1e3955558d9489a0e1b5604c0210802.

Makefile
jirutka: Resolved by 931a0b5038e2f9d869d7989c53b21f09509e6e3e.

jirutka:
I made the requested changes. Could you please take a second look?

lfittl (approved):

PR #128, opened 2022-02-18
fix(deparse): potential sigsegv

wesselvdv:
Fix potential sigsegv when passing an invalid protobuf (RawStmt without Stmt) for deparsing.

### DIFF
--- a/src/pg_query_deparse.c
+++ b/src/pg_query_deparse.c
@@ -2276,6 +2276,9 @@ static void deparseRangeVar(StringInfo str, RangeVar *range_var, DeparseNodeCont
 
 static void deparseRawStmt(StringInfo str, RawStmt *raw_stmt)
 {
+	if (raw_stmt->stmt == NULL)
+		elog(ERROR, "deparse error in deparseRawStmt: RawStmt with empty Stmt");
+	
 	deparseStmt(str, raw_stmt->stmt);
 }
 


src/pg_query_deparse.c
```diff
@@ -2276,7 +2276,12 @@ static void deparseRangeVar(StringInfo str, RangeVar *range_var, DeparseNodeCont
 
 static void deparseRawStmt(StringInfo str, RawStmt *raw_stmt)
 {
-	deparseStmt(str, raw_stmt->stmt);
+	if(raw_stmt->stmt != NULL) {
+		deparseStmt(str, raw_stmt->stmt);
+		return;
+	}
+
+	elog(ERROR, "deparse error in deparseRawStmt: RawStmt with empty Stmt");
```
lfittl: Whilst I generally like the idea of returning early, I think this makes it a bit confusing here - maybe we can write like this instead:

```c
if (raw_stmt->stmt == NULL)
  elog(ERROR, "deparse error in deparseRawStmt: RawStmt with empty Stmt");

deparseStmt(str, raw_stmt->stmt);
```

Note that `elog(ERROR, ...)` will raise an exception (or the C equivalent of it, i.e. unwind the stack), so you don't need to have a return in that if condition.

wesselvdv: Thanks for the review! I will update the PR with your suggestion.

lfittl:
@wesselvdv Merged - thanks for the contribution!

PR #136, opened 2022-04-29
Add operators used by ltree extension

svenklemm:
This makes libpg_query no longer error on the following operators
used by the ltree extension: ?<@ ?@> ?~ ?@ ^?


lfittl (approved):
:+1:

PR #154, opened 2022-10-30
Added SWIG Java changes

davidaventimiglia-professional:
Hello @lfittl !

I'm submitting this PR but I don't expect you to merge it.  Consider it more of a proposal.  Would you ever consider using something like [SWIG](https://www.swig.org/) to put language bindings directly into this project and library?  The reason I ask is that like there're bindings for Go, Ruby, etc., I wanted there to be bindings in Java.  Not seeing them, I decided to create them.  Being a lazy person, I picked an easy path, which was to use SWIG.  But, then I saw that it took little effort and I wondered if this might be a nice approach for laying down a common baseline foundation for Go, Ruby, etc.  as well.  Anyway, it's just a POC.  Were it to be taken forward, some things to do would be:

1. fix the broken build for protobuf-cpp + swig
2. add lots more test cases in Java
3. tighten up the Java/Maven build part
4. adopt [one-jar](https://one-jar.sourceforge.net/) approach
5. since Java is supposed to be cross-platform, cross-compile on many other architectures (Windows DLL, e.g.)

Anyway, if you get a chance take a look and please let me know what you think.  Thanks so much.
Regards,
David

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,11 @@ jobs:
         os: [ubuntu-20.04]
         compiler: [clang, gcc]
         protobuf_lib: [protobuf-c, protobuf-cpp]
-        valgrind: [valgrind,no-valgrind]
+        valgrind: [valgrind, no-valgrind]
+        swig: [java, none]      # NS: Add swig matrix.
+        exclude:                # NS: Excluding combinations (for now) that break the build.
+          - protobuf_lib: protobuf-cpp
+            swig: java
     steps:
     - name: Check out code
       uses: actions/checkout@v2
@@ -39,6 +43,9 @@ jobs:
     - name: Install Valgrind
       if: matrix.valgrind == 'valgrind'
       run: sudo apt update && sudo apt install -y valgrind
+    - name: Install SWIG        # NS: Install swig if needed.
+      if: matrix.swig != 'none'
+      run: sudo apt update && sudo apt install -y swig
     - name: Determine make flags
       id: make_flags
       run: |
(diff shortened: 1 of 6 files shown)


lfittl:
@davidaventimiglia Thanks for opening this PR for discussion!

Directly embedding other language bindings in the core library isn't necessarily off the table (and we already directly maintain both the Ruby and Rust bindings at pganalyze), but it would be a larger effort that is not currently prioritized on our own immediate roadmap (however there are other pg_query related priorities like updates to PG 14 and PG 15).

In any case I think it would entail (slightly) more than just using SWIG - for example, the idea behind using Protobufs for the parser format is to have a way to autogenerate node structs in different languages, and so we'd want to incorporate that when adding support for a new language.

Some libraries, like the Ruby library, also have additional helpers to access some of the data (e.g. to find all tables referenced in a query), although these helpers could also be maintained in the core library, it would be a bigger effort to rewrite that in C, and/or make that clear where its not supported.

And further, we may one day see a future where the parser library is shipped by Postgres itself (much like libpq), and if that ever were to happen, we would likely expect that language wrappers continue to be maintained outside of Postgres.

All that considered, for now I'm -0.5 on merging this PR, and would encourage you to instead write a separate Java wrapper library using SWIG. I think once that library has a certain state of completeness we could re-open this for discussion (and at that point I'll also review again the state of other popular wrapper libraries).

As a side note, on naming for a wrapper library, assuming you choose a license like MIT/BSD/Postgres, and we don't maintain a binding ourselves, we're cool with you naming such a wrapper library "pg_query_java", or similar, and giving attribution to the core library.

However, if you decide to use a copyleft license for the wrapper, we'd ask that you use a different name as a courtesy, to avoid confusion (that's why the Python wrapper is named "pglast", as its GPL licensed, and we asked the author to use a unique name to make sure the users of that library know to check whether the license is acceptable to their organization).

Thanks in any case for your interest & excited to see support for Java :)

davidaventimiglia-professional:
@lfittl that all makes sense.  Thanks for the detailed response!  I think what I'll do is the following.

1. Close this PR.  I got what i wanted, which was just your attention and thoughts on the subject.  :smiley: 
2. Move this material into a separate project with a compatible license (MIT/BSD/Postgres).  That's how it began life anyway.
3. Keep it at a very basic pure library level, meaning Java analogs of the additional Ruby helpers would go in _another_ project that consumes the lower-level pure library project.
4. Figure out what to do with the protobuffers.  Don't know much about them.  Time to learn!
5. Repeat the process for SWIG Python, since Python is the only other language I know well enough even to attempt this.
6. Refactor the SWIG Python + SWIG Python libraries for common patterns.
7. Write some blog posts and then close this chapter of my life.

Anyway, thanks so much!

lfittl:
@davidaventimiglia Nice, that sounds like a good plan :)

PR #155, opened 2022-11-02
Implement fingerprinting version 3.1

lfittl:
Consider lists that have a single NIL element as being present for determining
whether the parent field should be included in the fingerprint. In particular,
this matters for fingerprinting "SELECT DISTINCT" without columns specified.

Additionally, ignore more DDL attributes to avoid a lot of unique fingerprints:

(ListenStmt | UnlistenStmt | NotifyStmt).conditionname
=> This groups all LISTEN, UNLISTEN and NOTIFY statements together.
   Channels are often made unique to notify about different per end user
   objects, causing a lot of unique condition names.

CreateFunctionStmt.options and FunctionParameter.name
=> Some ORMs, like Sequelize.js, have a pattern where they repeatedely
   create functions with the exact name and types, but different function
   bodies -- this groups all these creations together. We may be more
   sophisticated in the future and introduce a "PL/pgSQL fingerprint".

DoStmt.args
=> This groups all anonymous DO code blocks together, since we don't
   have a good way to separate them otherwise, for now.

Additionally, fix a few normalization issues.


src/pg_query_fingerprint_defs.c
```diff
@@ -11639,6 +11602,13 @@ _fingerprintTableLikeClause(FingerprintContext *ctx, const TableLikeClause *node
     XXH3_freeState(prev);
   }
 
+  if (node->relationOid != 0) {
+    char buffer[50];
+    sprintf(buffer, "%d", node->relationOid);
+    _fingerprintString(ctx, "relationOid");
+    _fingerprintString(ctx, buffer);
+  }
```
msakrejda, line 11610: This adds the relation oid to the fingerprint if non-zero? Is this part of one of the changes in the description?

test/normalize_tests.c
```diff
@@ -17,6 +19,22 @@ const char* tests[] = {
...
+  "CREATE SUBSCRIPTION mysub CONNECTION 'host=192.168.1.50 port=5432 user=foo dbname=foodb' PUBLICATION mypublication, insert_only",
+  "CREATE SUBSCRIPTION mysub CONNECTION $1 PUBLICATION mypublication, insert_only",
+  "ALTER SUBSCRIPTION mysub SET PUBLICATION insert_only",
+  "ALTER SUBSCRIPTION mysub SET PUBLICATION insert_only",
+  "ALTER SUBSCRIPTION mysub CONNECTION 'host=192.168.1.50 port=5432 user=foo dbname=foodb'",
+  "ALTER SUBSCRIPTION mysub CONNECTION $1",
+  "CREATE USER MAPPING FOR bob SERVER foo OPTIONS (user 'bob', password 'secret')",
+  "CREATE USER MAPPING FOR bob SERVER foo OPTIONS (user $1, password $2)",
+  "ALTER USER MAPPING FOR bob SERVER foo OPTIONS (SET password 'public')",
+  "ALTER USER MAPPING FOR bob SERVER foo OPTIONS (SET password $1)",
```
msakrejda, line 37: Not blocking, but one might expect "mysub", "mypublication", "bob", and "foo" to be normalized in the above, no?

lfittl: Generally we've normalized constant values but not object references - and this follows that same intent.

src/pg_query_fingerprint_defs.c
lfittl: Hm, interesting. I wonder how that got in there. That seems more by accident than intentional. Will review (and likely remove). Thanks for noticing!

test/normalize_tests.c
msakrejda: Got it, makes sense.

seanlinsley (approved):

src/pg_query_fingerprint_defs.c
lfittl: Upon further review, this looks like it was missed in a prior update. The relationOid is always 0 in the raw parser, so this is effectively a no-op.

PR #166, opened 2022-11-20
GH actions: add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where libpg_query already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libpg_query'
+        dry-run: false
+        language: c
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libpg_query'
+        dry-run: false
+        language: c
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libpg_query'
+        fuzz-seconds: 300
```
lfittl, line 18: I just checked and looks like our GH actions already take ~15min due to valgrind being slow, so I think this should be acceptable to add.

lfittl (approved):
Let's give it a try - I think we may want to revise some of this further (i.e. run shorter if we see ourselves waiting a lot for slow fuzzings), but it seems like a good idea in general.

DavidKorczynski:
Awesome, thanks @lfittl -- naturally feel free to modify the action to whichever works the best for you. I'd be interested in knowing your preferences so will aim to check back again in some time!

PR #167, opened 2022-11-29
Upgrade to PSQL 15.1

msepga:
This upgrades `libpg_query` to Postgres 15.1. All tests have been updated, and tests were added for new constructs.

- `Boolean` nodes have been added to the parser, replacing usage of integers with the values `1`/`0`.
- In the postgres parser, `Value` was changed to be a union stored in the `A_Const` type. For this upgrade, we block `A_Const` from code generation, and instead manually implement the relevant required functions.
- The `COPY` statement deparser was also updated to attempt deparsing into the old Postgres 8.4-style syntax (e.g. `COPY foo FROM STDIN FREEZE CSV`). This ensures that the regression tests keep working, as new changes in the parse structures mean that these two syntaxes result in equivalent but slightly different parse trees.
- Support was added for `MERGE` statements.
- Support was added for more advanced publication objects, e.g. `CREATE PUBLICATION foo FOR TABLES IN SCHEMA CURRENT_SCHEMA`
- Support was added for `ALTER TABLE ALL IN TABLESPACE ...` statements


lelit:
Wow!

lelit:
Why `Boolean` is not present [here](https://github.com/pganalyze/libpg_query/blob/15-latest-dev/srcdata/struct_defs.json#L9085)?

msepga:
> Why `Boolean` is not present [here](https://github.com/pganalyze/libpg_query/blob/15-latest-dev/srcdata/struct_defs.json#L9085)?

Nice catch :) Fixed in [901ad3a](https://github.com/pganalyze/libpg_query/pull/167/commits/901ad3a327f356d13c9508affd698b3545929b3b)

lelit:
FYI, I almost completed upgrading pglast to this, and everything seems working great! :tada:

README.md
```diff
@@ -231,7 +231,8 @@ Each major version is maintained in a dedicated git branch. Only the latest Post
 
 | PostgreSQL Major Version | Branch     | Status              |
 |--------------------------|------------|---------------------|
-| 14                       | 14-latest  | Active development  |
+| 15                       | 15-latest  | Active development  |
+| 14                       | 14-latest  | Critical fixes only |
```
lfittl, line 235: :+1:

examples/simple.c
```diff
@@ -7,13 +7,12 @@
...
 #include <stdlib.h>
 
-size_t testCount = 7;
+size_t testCount = 6;
 const char* tests[] = {
   "SELECT 1",
   "SELECT * FROM x WHERE z = 2",
   "SELECT 5.41414",
   "SELECT $1",
-  "SELECT ?",
```
lfittl, line 16: In theory, should this also have been in the 14 branch? (no big deal if it isn't, just making sure I understand the change)

lfittl (approved):
Looks good overall!

For the fingerprint tests, can we also add one for MERGE that contains a constant value? Something like this:

```sql
MERGE INTO customer_account ca
USING (VALUES (1, 42), (2, 99)) t(customer_id, transaction_value)
ON t.customer_id = ca.customer_id
WHEN MATCHED THEN
  UPDATE SET balance = balance + transaction_value + 1
WHEN NOT MATCHED THEN
  INSERT (customer_id, balance)
  VALUES (t.customer_id, t.transaction_value);
```

examples/simple.c
msepga: Rebased against a commit fixing this in the `14-latest` branch

lfittl (approved):
:+1: Nicely done, and thanks for adding that extra fingerprint test!

Before merging it would be good to pull in the latest changes from 14-latest (really that should just be #168 which is now merged) - otherwise I think this is good to ship 🚢

PR #168, opened 2022-12-10
Ensure we don't rely on x86 and ARM intrinsics (i.e. SSE 4.2, etc)

lfittl:
Add three missing defines to always be unset to avoid this dependency:
- HAVE_X86_64_POPCNTQ
- USE_ARMV8_CRC32C
- USE_SSE42_CRC32C_WITH_RUNTIME_CHECK

Additionally, to help with generating the correct code on either ARM or x86, make sure we unset these defines before we run the extract source script. This now produces the identical output on x86 and ARM, besides some minor pg_config differences before the undefs at the end that undo those differences.

In passing, drop the "-msse4.2" flag passed in during source extraction, as we do not want to target SSE 4.2.


lfittl:
FWIW, this is mostly a cleanup step, since the relevant code parts are not actively used. Therefore we don't need to tag a new release for this (I think), but should make sure we also get this into 15-latest when we merge it.

msepga (approved):
LGTM :+1:

Out of curiosity: Given the ubiquity of SSE 4.2 support, what made this necessary?

lfittl:
> Out of curiosity: Given the ubiquity of SSE 4.2 support, what made this necessary?

In practice the `-msse4.2` switch to clang doesn't make a difference here (the extracted code is the same) - I removed it since the intent of the extraction script is to generate something that's somewhat architecture independent (i.e. doesn't pull in code that's x86-only, etc.), and the SSE instruction set is by definition only available on x86 / Intel-based platforms.

PR #171, opened 2023-01-03
Extend PgQueryProtobuf.len to machine word size

### LINKED ISSUE #114: deparse_protobuf() broken on big-endian
Hi,
as can be seen on https://buildd.debian.org/status/logs.php?pkg=pglast&ver=5.0%7Edev0-1, pglast is broken on big-endian architectures. (Sorry for the version number messup, I had uploaded a git snapshot of the v5 branch as 5.0~dev0 a week or two before you released this version. Thanks for the release and the fixes in there!)

The specific failure is this:

```
=================================== FAILURES ===================================
____________________________ test_deparse_protobuf _____________________________

    def test_deparse_protobuf():
>       assert deparse_protobuf(parse_sql_protobuf('select 1')) == 'SELECT 1'
E       AssertionError: assert '' == 'SELECT 1'
E         - SELECT 1

tests/test_parser.py:184: AssertionError
```

The above log is from the "Debian" dev0 package, but I can reproduce the problem on the "real" dev0 version as well:

```
>>> from pglast.parser import deparse_protobuf, parse_sql_protobuf
>>> deparse_protobuf(parse_sql_protobuf('select 1'))
''
>>> parse_sql_protobuf('select 1')
b'\x08\xf1\x93\t\x12\x1c\n\x1a\xd2\x03\x17\x1a\x10\x9a\x0b\r\x1a\t\xf2\x0e\x06\n\x02\x08\x01X\x07 \x07p\x01\x88\x01\x01'
>>> deparse_protobuf(parse_sql_protobuf('select'))
''
```

df7cb:
The len field is expected to be size_t from the protobuf side (see the definition of pg_query__parse_result__unpack()). Also, clients are expecting to be able to use size_t-like types when filling the PgQueryProtobuf structure.

This is how this problem was discovered in pglast: deparse_protobuf(parse_sql_protobuf('select 1')) would always return '' on big-endian machines.

deparse_protobuf() populated a 64-bit machine-word-sized len field, but libpg-query would only read an "int" from it, which is 32-bit only even on 64-bit machines. On big-endian machines, the "wrong" half of the integer is read, so the len is effectively always zeros.

Close lelit/pglast#114.

### DIFF
--- a/examples/scan.c
+++ b/examples/scan.c
@@ -39,7 +39,7 @@ int main() {
     } else {
       scan_result = pg_query__scan_result__unpack(NULL, result.pbuf.len, (void *) result.pbuf.data);
 
-      printf("  version: %d, tokens: %ld, size: %d\n", scan_result->version, scan_result->n_tokens, result.pbuf.len);
+      printf("  version: %d, tokens: %ld, size: %zu\n", scan_result->version, scan_result->n_tokens, result.pbuf.len);
       for (j = 0; j < scan_result->n_tokens; j++) {
         scan_token = scan_result->tokens[j];
         token_kind = protobuf_c_enum_descriptor_get_value(&pg_query__token__descriptor, scan_token->token);
--- a/pg_query.h
+++ b/pg_query.h
@@ -2,6 +2,7 @@
 #define PG_QUERY_H
 
 #include <stdint.h>
+#include <sys/types.h>
 
 typedef struct {
 	char* message; // exception message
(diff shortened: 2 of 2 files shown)


lfittl (approved):
Looks good, thanks for the contribution! - assuming tests can pass we can merge it :)

df7cb:
Thanks!

Fwiw the issue was on 64-bit big endian (because `int` is still 32-bit only even on 64-bits).

PR #172, opened 2023-01-19
pg_config.h: NetBSD support

bsiegert:
This is the same as pganalyze/pg_query_go#72 except one level higher.

NetBSD has `strchrnul` too but with a slightly different prototype. Without this change, there is a compiler error for a conflicting declaration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,8 +146,8 @@ $(PGDIR):
 	# Avoid CRC extension usage to ensure we are not architecture-dependent
 	echo "#undef USE_ARMV8_CRC32C" >> $(PGDIR)/src/include/pg_config.h
 	echo "#undef USE_SSE42_CRC32C_WITH_RUNTIME_CHECK" >> $(PGDIR)/src/include/pg_config.h
-	# Ensure we don't fail on systems that have strchrnul support (FreeBSD)
-	echo "#ifdef __FreeBSD__" >> $(PGDIR)/src/include/pg_config.h
+	# Ensure we don't fail on systems that have strchrnul support (FreeBSD and NetBSD)
+	echo "#if defined(__FreeBSD__) || defined(__NetBSD__)" >> $(PGDIR)/src/include/pg_config.h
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -146,8 +146,8 @@ $(PGDIR):
 	# Avoid CRC extension usage to ensure we are not architecture-dependent
 	echo "#undef USE_ARMV8_CRC32C" >> $(PGDIR)/src/include/pg_config.h
 	echo "#undef USE_SSE42_CRC32C_WITH_RUNTIME_CHECK" >> $(PGDIR)/src/include/pg_config.h
-	# Ensure we don't fail on systems that have strchrnul support (FreeBSD)
-	echo "#ifdef __FreeBSD__" >> $(PGDIR)/src/include/pg_config.h
+	# Ensure we don't fail on systems that have strchrnul support (FreeBSD and NetBSD)
+	echo "#if defined(__FreeBSD__) || defined(__NetBSD__)" >> $(PGDIR)/src/include/pg_config.h
```
lfittl, line 150: We do actually also need to apply it to the generated pg_config.h file in the repo (we cache the result of the Makefile and only run extract_source as needed) -- you can just make the edit directly in the file, or run the `extract_source` locally to generate it (but that might be a bit more involved):

https://github.com/pganalyze/libpg_query/blob/15-latest/src/postgres/include/pg_config.h#L1035

lfittl (commented):
Thanks for getting this change in here!

There is actually one more small detail (I wasn't fully clear in my initial ask I realize), once that's fixed we can merge it :)

Makefile
bsiegert: Done. I did not manage to get `make extract_source` to work (neither on NetBSD nor on Debian) so this is a manual edit. PTAL :)

lfittl (approved):

bsiegert:
Can someone merge this, because I cannot :)

lfittl:
> Can someone merge this, because I cannot :)

Looks like there was a seemingly random build failure - will merge after thats resolved (hopefully on a re-run, that I started just now).

PR #174, opened 2023-02-02
License file's copyright is stuck in 2015

yrashk:
Update it to reflect libpog_query's active status.

P.S. I am incorporating portions of the library into another open source project (more on that later) and I noticed that the copyright in the license hasn't been updated in a while.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015, Lukas Fittl <lukas@fittl.com>
+Copyright (c) 2015-2023, Lukas Fittl <lukas@fittl.com>
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without


lfittl:
@yrashk Thanks for the contribution! You're right that it'd technically be more correct to list the years here. Let me actually update this to reflect our company name, since its more correct nowadays, as its no longer just me (e.g. last release was driven by @msepga)

seanlinsley:
FWIW: https://stackoverflow.com/questions/2390230

lfittl:
> FWIW: https://stackoverflow.com/questions/2390230

Thanks for pointing that out - I think for libpg_query in particular we can follow what the Postgres project does (and in fact, we should reflect their copyright years correctly), see their annual copyright notice update: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=fdaba0a2d749bff497313b584f86855a1440d43e

See https://github.com/pganalyze/libpg_query/pull/175 which does just that.

lfittl:
Fixed via #175

PR #175, opened 2023-02-02
Update copyright notice to reflect current years + pganalyze as author

lfittl:
Since this library is now collectively maintained by pganalyze as a company, not just by myself (Lukas), it seems logical to expand the LICENSE file to cover this copyright explicitly.


lfittl:
FWIW, as pointed out by Sean in https://github.com/pganalyze/libpg_query/pull/174#issuecomment-1414190490, one generally does not need to do this in the US, as you can't just magically expand your copyright by updating a date.

That said, I think for this library it seems reasonable follow Postgres project standard, which is to update the copyright years.

(this does not constitute a legal opinion, and we may change this in the future if advised differently by legal counsel)

msepga (approved):

PR #177, opened 2023-02-06
Allow trailing junk in numeric literals

msepga:
This reverts PSQL commit [2549f066](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=2549f0661bd28571d7200d6f82f752a7ee5d47e1), a grammar change which rejects trailing text on numeric literals. The original patch was intended to support future numeric literal syntax, however the changes did not need to be so aggressive. For now this is reverted, until a more relaxed grammar is implemented, with the upcoming PG16 release in mind.


test/parse_tests.c
```diff
@@ -65,6 +65,8 @@ const char* tests[] = {
   "{\"version\":150001,\"stmts\":[{\"stmt\":{\"AlterTableMoveAllStmt\":{\"orig_tablespacename\":\"foo\",\"objtype\":\"OBJECT_TABLE\",\"roles\":[{\"RoleSpec\":{\"roletype\":\"ROLESPEC_CSTRING\",\"rolename\":\"bar\",\"location\":43}},{\"RoleSpec\":{\"roletype\":\"ROLESPEC_CSTRING\",\"rolename\":\"quux\",\"location\":48}}],\"new_tablespacename\":\"fred\",\"nowait\":true}}}]}",
   "CREATE PUBLICATION foo FOR TABLES IN SCHEMA bar",
   "{\"version\":150001,\"stmts\":[{\"stmt\":{\"CreatePublicationStmt\":{\"pubname\":\"foo\",\"pubobjects\":[{\"PublicationObjSpec\":{\"pubobjtype\":\"PUBLICATIONOBJ_TABLES_IN_SCHEMA\",\"name\":\"bar\",\"location\":44}}]}}}]}",
+  "SELECT 123abc",
```
lfittl, line 68: +1 to adding a test here, I wonder if we should use an example that includes param refs as well? (or just have that)

Anticipating that in Postgres 16 we'd not support this syntax for regular integers, but would keep supporting the syntax for param refs.

msepga: Have added a test for `SELECT 123 AS abc where 456=$1OR 10=11` 👍

lfittl (approved):
:+1:

lfittl:
For additional context on anyone else looking at this PR in the future: The main problem that libpg_query users have with this upstream change (and that we've seen in the pganalyze app), is that pg_stat_statements can produce output like "SELECT * FROM x WHERE y=$1ORz=$2" (note the missing space between `$1` and `OR`), which is rejected by the Postgres 15 parser.

The upstream change here was mainly intended to allow for `0x123def` and `100_000_000` syntax changes in Postgres 16, and didn't really intentionally change the parameter reference handling, from what I can gather. For Postgres 16 we'll try to make a more surgical change that only keeps the old behavior (allowing trailing "junk") for parameter references, whilst supporting the new syntax - but for PG15 reverting that preparatory commit (as done here), is the easier change.

PR #182, opened 2023-02-26
Deparser: Ensure index names are quoted as identifiers

### LINKED ISSUE #77: Deparse of index with period (.) in name leads to syntax error
Hi,

If I parse and then de-parse the following statement:

```sql
CREATE INDEX "foo.index" ON foo USING btree (bar)
```

then I end up with a result where the index name is unquoted, which is a syntax error:

```sql
CREATE INDEX foo.index ON foo USING btree (bar)
```

PostgresQL (tested on 10.23 and 15.1) reports:

```
postgres=# CREATE INDEX foo.index ON foo USING btree (bar);
ERROR: syntax error at or near "." (SQLSTATE 42601)
```

lfittl:
Fixes https://github.com/pganalyze/pg_query_go/issues/77

### DIFF
--- a/src/pg_query_deparse.c
+++ b/src/pg_query_deparse.c
@@ -7938,7 +7938,7 @@ static void deparseIndexStmt(StringInfo str, IndexStmt *index_stmt)
 
 	if (index_stmt->idxname != NULL)
 	{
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
🍰

PR #183, opened 2023-02-26
Move deparser to dedicated file without pg_query dependencies

### LINKED ISSUE #180: Making query deparsing indepenently usable
I am using `libpg_query`'s deparsing functionality in [Omnigres](https://github.com/omnigres/omnigres/tree/master/extensions/omni_sql). However, since it's primarily a collection of extensions, the primary use case of libpg_query itself doesn't quite fit. I don't use it independently of Postgres, I am supplementing Postgres with the deparsing functionality when I need to manipulate or analyze queries.

I've done a fairly simple job integrating deparsing code (as seen above) but going forward, every time deparsing changes, I'd need to repeat a few steps, which is not ideal:

* Getting rid of pg_query_* API related stuff as it is not relevant
* Exposing `deparseRawStmt` function

I wonder if you'd be open to a patch that, for example, makes deparsing turn off pg_query_ and expose `deparseRawStmt` conditionally (perhaps opt-in to avoid the exposure of this throughout libpg_query itself)?

This way I can take the deparsing code and simply compile it with `-DISOLATED_PGQUERY_DEPARSING` and updating this dependency will be as simple as grabbing a new version of the file.

Happy to prepare a PR if this sounds like a good start.

lfittl:
This enables external projects that are interested to work with a raw parsetree to use the libpg_query deparser, without pulling in the main pg_query headers.

Fixes #180.


lfittl:
Note that unfortunately git doesn't automatically detect this as a file move (since the old file still exists), but there are no changes in the deparser code, except for making deparseRawStmt exported, as is more easily seen by doing something like:

```
git diff 15-latest:src/pg_query_deparse.c src/postgres_deparse.c
```

keiko713:
Looks like you need to adopt changes you made with `src/pg_query_deparse.c`?

yrashk:
Following up on this from #180 (sorry for the delay, been busy). Do you plan to update & merge this in?

yrashk:
Bump?

yrashk:
Anything I can do to help with this change? I can try and send new PR, rebased off the latest master.

lfittl:
> Anything I can do to help with this change? I can try and send new PR, rebased off the latest master.

Yes, that would actually help - this is on my list to get to, but if you make a new PR that is rebased it makes it an easy merge :)

yrashk:
@lfittl I hope I captured the essence of the change correctly in #192

lfittl:
Superseded by #192, which is now merged.

PR #203, opened 2023-08-04
Fix builds on glibc >= 2.38

### LINKED ISSUE #202: Build failure with glibc 2.38 due to strchrnul being available by default
Per https://github.com/pganalyze/pg_query_go/issues/95 and https://github.com/pganalyze/pg_query/issues/298, the `strchrnul` function is now always available in glibc 2.38+, even if _GNU_SOURCE is not set (see [upstream commit](https://sourceware.org/git/?p=glibc.git;a=commit;h=30e3ca78f92be44f46851557f54b47e789bd3416)).

This conflicts with Postgres' built-in definition, which we currently use on all non-BSD systems.

The likely fix here is to broaden the condition we put in pg_config.h to also check for glibc 2.38+ on any OS: https://github.com/pganalyze/libpg_query/blob/cb3aa9ee69f0bc2d722e68019059fa1680c937b0/Makefile#L150

(i.e. something like `((__GLIBC__ == 2 && __GLIBC_MINOR__ >= 38) || __GLIBC__ > 2)`, inspired by the proposed patch in the  bug report for the Ruby library)

msepga:
`glibc` now defines `strchrnul`, so we add a check to see if `glibc` is being used, and instead import the symbol if so.

Closes #202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All versions are tagged by the major Postgres version, plus an individual semver for this library itself.
 
+## 15-4.2.3   2023-07-07
+
+* Fix builds when compiling with `glibc >=  2.38` [#203](https://github.com/pganalyze/libpg_query/pull/203)
+
 ## 15-4.2.2   2023-07-07
 
 * Deparser:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All versions are tagged by the major Postgres version, plus an individual semver
 ## 15-4.2.3   2023-07-07
 
 * Fix builds when compiling with `glibc >=  2.38` [#203](https://github.com/pganalyze/libpg_query/pull/203)
+* Deparser: Add support for COALESCE and other expressions in LIMIT clause [#199](https://github.com/pganalyze/libpg_query/pull/199)
 
 ## 15-4.2.2   2023-07-07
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All versions are tagged by the major Postgres version, plus an individual semver for this library itself.
 
+## 15-4.2.3   2023-07-07
+
+* Fix builds when compiling with `glibc >=  2.38` [#203](https://github.com/pganalyze/libpg_query/pull/203)
```
lfittl, line 7: Would be good to also add a changelog entry for #199, like:

```
* Deparser: Add support for COALESCE and other expressions in LIMIT clause
```

lfittl (approved):

src/postgres/include/pg_config.h
```diff
@@ -1032,6 +1032,7 @@
 #undef HAVE__GET_CPUID
 #undef USE_ARMV8_CRC32C
 #undef USE_SSE42_CRC32C_WITH_RUNTIME_CHECK
-#if defined(__FreeBSD__) || defined(__NetBSD__)
+#include <features.h>
```
lfittl: This doesn't exist for me on macOS - do we actually need this?

lfittl: (resolved after the last commit)

PR #210, opened 2023-09-30
Makefile: Use -Wall for compiling tests

evanj:
CFLAGS uses -Wall along with some -Wno-* options. However, tests are compiled using TEST_CFLAGS, so they did not use these options. Compile tests with -Wall. This also reorders the TEST_CFLAGS options to match the CFLAGS options above for easier comparison.

test/concurrency: Fix unused arg warning from -Wall. Sort includes.

Feel free to reject this pull request since it doesnt really change anything. I just noticed it while making another change to the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ override CFLAGS += -g -I. -I./vendor -I./src/postgres/include -Wall -Wno-unused-
 
 override PG_CONFIGURE_FLAGS += -q --without-readline --without-zlib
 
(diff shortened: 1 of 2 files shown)


lfittl (approved):
:+1: Seems reasonable, thanks for the contribution!

Note that we intentionally don't check for some warnings for the main library code (since Postgres itself isn't warning free), but the compilation flags for those units uses `CFLAGS`, not `TEST_CFLAGS`, so we are good in that regard.

Issue #50, opened 2018-11-10
Support configuring the parser w/ relevant server settings?


ethanresnick:
My understanding is that some PG server options can change how parsing works (e.g., `allow_nulls` and `standard_conforming_strings`). Does libpg_query support configuring the parser with these settings, to match the server? Can/should it?

lfittl:
@ethanresnick Today there is no mechanism to modify these settings, mostly since there are very limited use cases for setting them.

Do you have a reason to want to set them / what's the use case?

ethanresnick:
> Do you have a reason to want to set them

No, I was mostly just curious; my thought was that a user of this library who did have those settings applied on their server would likely want to match them in libpq_query, as it seems like having a guarantee of consistent parsing between the server and wherever libpq_query is running could be an important security property, and part of the reason people would want to use this library

jawj:
+1 for `standard_conforming_strings`. If you want to parse like pg, you need to be able to match pg's alternative parsing modes.

lfittl:
From a quick look, it looks like supporting `standard_conforming_strings` (as well as `backslash_quote` and `escape_string_warning`) should be pretty straightforward, since we already have thread local variables that represent their state: https://github.com/pganalyze/libpg_query/blob/15-latest/src/postgres/include/parser/parser.h#L58

Note that `allow_nulls` isn't relevant to raw parsing in my understanding - its only applicable once the input data is parsed, and hence would not matter for pg_query's use case (which is to work with raw parse trees).

Since this is mainly about the API then (if you use the library in C, you could already set those variables before calling pg_query_parse), I'm curious how anyone interested in this is using pg_query today, i.e. from which wrapper/for which functions.

@ethanresnick @jawj Could you share a few notes which functions you are calling where you'd see a benefit from that? (e.g. are you using the Ruby wrapper library to call the parser function itself, or the fingerprint functions, etc)

ethanresnick:
@lfittl It's been quite a while since I worked with this library, but, at the time I opened the issue, I was using [my own fork](https://github.com/ethanresnick/node-pg-query-native) of a then-current Javascript wrapper, and calling its `parseQuery`/`parseQuerySync` methods. I was not using query fingerprinting. I was also using PL/PGSQL parsing, in case these settings effect that.

jawj:
I'm using it from Node — `import { parse } from 'pgsql-parser';` — to parse out separate SQL statements in a text block.

jawj:
Thank you!

PR #216, opened 2023-10-28
Support passing parse mode to parse/fingerprint functions

### LINKED ISSUE #215: Parsing queries from plpgsql
Trying to parse queries found in PLpgSQL_expr after calling pg_query_parse_plpgsql. In version 14 and later postgres changed how it returns the query in the struct. Using the examples below, before it would return `SELECT EXISTS(SELECT 1 FROM sample WHERE name = n)` but now it returns the original query `EXISTS(SELECT 1 FROM sample WHERE name = n)` with a parseMode field. Any issues/concerns with exposing parseMode and being able to pass it to parse functions? 

```sql
CREATE OR REPLACE FUNCTION public.function_a(n TEXT) 
RETURNS boolean LANGUAGE plpgsql 
AS 
$$ 
BEGIN 
    IF EXISTS(SELECT 1 FROM sample WHERE name = n) THEN
        RETURN true;
    END IF;
    RETURN false;
END;
$$;
```

#### libpg_query on 13-latest branch
```json
[
    {
        "PLpgSQL_function": {
            "datums": [
                {
                    "PLpgSQL_var": {
                        "refname": "n",
                        "datatype": {
                            "PLpgSQL_type": {
                                "typname": "UNKNOWN"
                            }
                        }
                    }
                },
                {
                    "PLpgSQL_var": {
                        "refname": "found",
                        "datatype": {
                            "PLpgSQL_type": {
                                "typname": "UNKNOWN"
                            }
                        }
                    }
                }
            ],
            "action": {
                "PLpgSQL_stmt_block": {
                    "lineno": 1,
                    "body": [
                        {
                            "PLpgSQL_stmt_if": {
                                "lineno": 1,
                                "cond": {
                                    "PLpgSQL_expr": {
                                        "query": "SELECT EXISTS(SELECT 1 FROM sample WHERE name = n)"
                                    }
                                },
                                "then_body": [
                                    {
                                        "PLpgSQL_stmt_return": {
                                            "lineno": 1,
                                            "expr": {
                                                "PLpgSQL_expr": {
                                                    "query": "SELECT true"
                                                }
                                            }
                                        }
                                    }
                                ]
                            }
                        },
                        {
                            "PLpgSQL_stmt_return": {
                                "lineno": 1,
                                "expr": {
                                    "PLpgSQL_expr": {
                                        "query": "SELECT false"
                                    }
                                }
                            }
                        }
                    ]
                }
            }
        }
    }
]
```

#### libpg_query on 15-latest branch
```json
[
    {
        "PLpgSQL_function": {
            "datums": [
                {
                    "PLpgSQL_var": {
                        "refname": "n",
                        "datatype": {
                            "PLpgSQL_type": {
                                "typname": "UNKNOWN"
                            }
                        }
                    }
                },
                {
                    "PLpgSQL_var": {
                        "refname": "found",
                        "datatype": {
                            "PLpgSQL_type": {
                                "typname": "UNKNOWN"
                            }
                        }
                    }
                }
            ],
            "action": {
                "PLpgSQL_stmt_block": {
                    "lineno": 1,
                    "body": [
                        {
                            "PLpgSQL_stmt_if": {
                                "lineno": 1,
                                "cond": {
                                    "PLpgSQL_expr": {
                                        "query": "EXISTS(SELECT 1 FROM sample WHERE name = n)"
                                    }
                                },
                                "then_body": [
                                    {
                                        "PLpgSQL_stmt_return": {
                                            "lineno": 1,
                                            "expr": {
                                                "PLpgSQL_expr": {
                                                    "query": "true"
                                                }
                                            }
                                        }
                                    }
                                ]
                            }
                        },
                        {
                            "PLpgSQL_stmt_return": {
                                "lineno": 1,
                                "expr": {
                                    "PLpgSQL_expr": {
                                        "query": "false"
                                    }
                                }
                            }
                        }
                    ]
                }
            }
        }
    }
]
```

lcheruka:
Fixes #215


src/pg_query_parse.c
```diff
@@ -42,7 +42,7 @@ PgQueryInternalParsetreeAndError pg_query_raw_parse(const char* input)
 
 	PG_TRY();
 	{
-		result.tree = raw_parser(input, RAW_PARSE_DEFAULT);
+		result.tree = raw_parser(input, mode);
```
seanlinsley: I'm not sure what the fix should be, but there's a warning here:
```
compiling src/pg_query_parse.c
src/pg_query_parse.c:45:35: warning: implicit conversion from enumeration type 'PgQueryParseMode' to different enumeration type 'RawParseMode' [-Wenum-conversion]
                result.tree = raw_parser(input, mode);
                              ~~~~~~~~~~        ^~~~
```

test/fingerprint_opts.c
```diff
@@ -0,0 +1,54 @@
...
+		}
+		else if (strcmp(result.fingerprint_str, tests[i + 2]) == 0)
+		{
+			printf(".");
+		}
+		else
+		{
+			ret_code = -1;
+			printf("INVALID result for \"%s\" with %d mode\nexpected: \"%s\"\nactual: \"%s\"\nactual tokens: ", tests[i], mode, tests[i + 2], result.fingerprint_str);
+			pg_query_fingerprint_with_opts(tests[i], mode, true);
```
seanlinsley, line 35: It looks like the `actual tokens: ` should be removed, as well as this `pg_query_fingerprint_with_opts` call. Or did you intend to do something more here?

lcheruka: This was copied from the existing fingerprint.c. I believe the second call prints the actual tokens.

lfittl:
Mainly commenting so I don't forget:

For the newly introduced `_opt` functions, I wonder if we should also consider the use cases requested in https://github.com/pganalyze/libpg_query/issues/50 here, i.e. to set GUCs that affect parsing.

Need to think through this more how it could be part of the API, but might be worth doing now so we don't have to break the `_opt` function API in the future.

lfittl:
> Need to think through this more how it could be part of the API, but might be worth doing now so we don't have to break the `_opt` function API in the future.

I've gone ahead and done that change now on this branch (thanks for allowing maintainer edits!), by utilizing an integer as a bitmask instead of a simple enum. From a usability perspective this is essentially the same as if you just care about the parse mode, but if you want to set the options you can just use bitwise OR to enable them.

Overall I think I'm happy with this now. @seanlinsley might be worth if you take another look to see if there are any concerns with that last commit, but otherwise I think we can merge.

lfittl (approved):

msepga (approved):

seanlinsley (approved):

lfittl:
@lcheruka Thanks for the contribution! Merged with slight revisions to allow passing additional options beyond the parse mode.

PR #217, opened 2023-11-03
Fix: ar variable in Makefile to inherit upstream ar command

gregnr:
Fantastic library!

I'm working on compiling this library to WASM using Emscripten (specifically for [libpg-query-node](https://github.com/pyramation/libpg-query-node)). Emscripten has a command `emmake` that simply wraps `make` with `CC`, `CXX`, `AR`, etc set to the Emscripten drop-in replacements (`emcc`, `em++`, `emar`).

Surprisingly running `emmake make` on this repository's `Makefile` JustWorked™ with the exception of a one-line change to the `Makefile` which is what this PR proposes.

We just need to replace the hardcoded `AR` variable:
```
AR = ar rs
```

with a reference to the upstream inherited `AR` variable:

```
AR := $(AR) rs
```

so that Emscripten's `emar` replacement is used.

Let me know if you have any questions or need any more clarification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ CLEANLIBS = $(ARLIB)
 CLEANOBJS = $(OBJ_FILES)
 CLEANFILES = $(PGDIRBZ2)
 
+AR ?= ar
 AR := $(AR) rs
 INSTALL = install
 LN_S = ln -s
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ CLEANLIBS = $(ARLIB)
 CLEANOBJS = $(OBJ_FILES)
 CLEANFILES = $(PGDIRBZ2)
 
-AR = ar rs
+AR := $(AR) rs
 INSTALL = install
 LN_S = ln -s
 RM = rm -f


Makefile
```diff
@@ -62,7 +62,7 @@ CLEANLIBS = $(ARLIB)
 CLEANOBJS = $(OBJ_FILES)
 CLEANFILES = $(PGDIRBZ2)
 
-AR = ar rs
+AR := $(AR) rs
```
lfittl, line 66: Do we need to treat this like we treat CC below, which is to default it to `ar` explicitly?

i.e. make it

```
AR ?= ar
AR := $(AR) rs
```

lfittl:
@gregnr Thanks for the contribution! I think that looks good, with one minor question inline.

Makefile
gregnr: I suppose it doesn't hurt! I left the first part out based on these `Makefile` docs:
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

But not 100% if those defaults work in all environments. Will update with your suggestion.

gregnr:
@lfittl thanks for the quick review! I've pushed an update based on your suggestion.

PS. are you open to applying this change retroactively to v13 and v14 branches for backwards support? If so I can create 2 more PR's for those.

lfittl (approved):

lfittl:
> PS. are you open to applying this change retroactively to v13 and v14 branches for backwards support? If so I can create 2 more PR's for those.

We generally don't back patch, unless there is a good reason to do so. But not against it per se. Do you have a use case that requires using older parser versions?

gregnr:
Thanks for merging!

> Do you have a use case that requires using older parser versions?

Yes - I was planning to support v13 and v14 WASM builds as well (when paired with older Postgres versions), for which I would reference the respective version tags on this repo ([v13](https://github.com/pganalyze/libpg_query/tree/13-latest), [v14](https://github.com/pganalyze/libpg_query/tree/14-latest)). But they will need the same `Makefile` patch to work.

lfittl:
> > Do you have a use case that requires using older parser versions?
> 
> Yes - I was planning to support v13 and v14 WASM builds as well (when paired with older Postgres versions), for which I would reference the respective version tags on this repo ([v13](https://github.com/pganalyze/libpg_query/tree/13-latest), [v14](https://github.com/pganalyze/libpg_query/tree/14-latest)). But they will need the same `Makefile` patch to work.

Ack, makes sense. I think we can support that, if you want to make a PR for them :)

gregnr:
> Ack, makes sense. I think we can support that, if you want to make a PR for them :)

I finally got around to doing this. Here are the PR's for v13 and v14:

- **v13:** https://github.com/pganalyze/libpg_query/pull/239
- **v14:** https://github.com/pganalyze/libpg_query/pull/240

Thanks @lfittl!

lfittl:
> > Ack, makes sense. I think we can support that, if you want to make a PR for them :)
> 
> I finally got around to doing this. Here are the PR's for v13 and v14:
> 
>     * **v13:** [Fix: ar variable in Makefile (back patch to v13) #239](https://github.com/pganalyze/libpg_query/pull/239)
> 
>     * **v14:** [Fix: ar variable in Makefile (back patch to v14) #240](https://github.com/pganalyze/libpg_query/pull/240)
> 
> 
> Thanks @lfittl!

@gregnr Great, thanks - merged both!

Did you need tagged patch releases for your use case, or is having them in the respective branches sufficient?

gregnr:
Thanks @lfittl!

A tagged release for v13 and v14 would be amazing if possible. Otherwise I can always reference a commit instead.

Issue #169, opened 2022-12-13
Compilation error reported


nIxedoahz:
Env:
     protobuf-c 1.1.1
     libprotoc 2.6.1
     libpg_query-14-3.0.0
     gcc version 4.8.1

error:(make)
compiling src/pg_query_split.c
In file included from ./src/postgres/include/postgres.h:47:0,
                 from src/pg_query_internal.h:4,
                 from src/pg_query_split.c:2:
./src/postgres/include/utils/elog.h:358:1: error: unknown type name 'sigjmp_buf'
 extern PGDLLIMPORT __thread  sigjmp_buf *PG_exception_stack;
 ^
src/pg_query_split.c: In function 'pg_query_split_with_scanner':
./src/postgres/include/utils/elog.h:315:3: error: unknown type name 'sigjmp_buf'
   sigjmp_buf *_save_exception_stack = PG_exception_stack; \
   ^
src/pg_query_split.c:54:3: note: in expansion of macro 'PG_TRY'
   PG_TRY();
   ^
./src/postgres/include/utils/elog.h:317:3: error: unknown type name 'sigjmp_buf'
   sigjmp_buf _local_sigjmp_buf; \
   ^
src/pg_query_split.c:54:3: note: in expansion of macro 'PG_TRY'
   PG_TRY();
   ^
src/pg_query_split.c:54:3: warning: implicit declaration of function 'sigsetjmp' [-Wimplicit-function-declaration]
src/pg_query_split.c:129:5: warning: implicit declaration of function 'strdup' [-Wimplicit-function-declaration]
     result.stderr_buffer = strdup(stderr_buffer);
     ^
src/pg_query_split.c:129:26: warning: assignment makes pointer from integer without a cast [enabled by default]
     result.stderr_buffer = strdup(stderr_buffer);
                          ^
src/pg_query_split.c:141:22: warning: assignment makes pointer from integer without a cast [enabled by default]
     error->message   = strdup(error_data->message);
                      ^
src/pg_query_split.c:142:22: warning: assignment makes pointer from integer without a cast [enabled by default]
     error->filename  = strdup(error_data->filename);
                      ^
src/pg_query_split.c:143:22: warning: assignment makes pointer from integer without a cast [enabled by default]
     error->funcname  = strdup(error_data->funcname);
                      ^
make: *** [src/pg_query_split.o] Error 1

lfittl:
@nIxedoahz Thanks for the report!

[...]

PR #221, opened 2023-11-23
Use "-std=gnu99" by default

SunBeau:
I get the following error on centos7

get another error as follows, if set -std=c99

it will be ok, if set -std=gnu99.
so i think that it should use "-std=gnu99" by default.
Related question: https://github.com/pganalyze/libpg_query/issues/169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ override PG_CONFIGURE_FLAGS += -q --without-readline --without-zlib
 override TEST_CFLAGS += -g -I. -I./vendor -Wall
 override TEST_LDFLAGS += -pthread
 
+override CFLAGS += -std=gnu99
+override TEST_CFLAGS += -std=gnu99
+
 CFLAGS_OPT_LEVEL = -O3
 ifeq ($(DEBUG),1)
 	CFLAGS_OPT_LEVEL = -O0


lfittl:
Thanks for the contribution!

In my understanding, this mainly helps older compilers that don't default to C99 already - so I think that generally sounds good. And per my reading of the Postgres mailing list archives, upstream requires C99 since Postgres 12, so we're good in that regard.

That said, I do wonder why `-std=c99` doesn't work - that would be slightly preferable to `-std=gnu99` if we can make it work. I'm not sure why it errors out like in your second screenshot - maybe we're just missing an include header that GNU mode automatically pulls in?

duckinator:
I happened to notice the title of this PR today, and was curious why gnu99 was suggested.

Turns out the explanation for these errors is that `strdup` and `sigsetjmp`/`sigjmp_buf` are part of the _POSIX_ standard, not the C standard.

Adding `-D_DEFAULT_SOURCE` to `CFLAGS` will cause gcc (and I believe clang) to expose POSIX functionality.

In addition, `typeof` (which the codebase uses) was originally a GNU extension, but has became standard in C23.

So, everything builds if you add `-std=c23 -D_DEFAULT_SOURCE` to CFLAGS.

PR #222, opened 2023-11-29
Fix deparse `ALTER TABLE table_name ENABLE TRIGGER ALL` syntax errors

CDThomas:
This change fixes syntax errors output by parsing/deparsing `ALTER TABLE table_name ENABLE TRIGGER ALL` statements. Previously the deparsed statement would be missing the `ALL` keyword.

Looks like the bug was a typo in the deparse logic that caused the `ALL` keyword to be left off.

This change adds the `ALL` keyword to the output string when `ENABLE TRIGGER ALL` is used. This change also adds coverage for other forms of `ALTER TABLE` statments that use `ENABLE`/`DISABLE` `TRIGGER`, but the other cases look to already be handled correctly.


lfittl:
Good catch - thanks for the contribution!

lfittl (approved):

PR #223, opened 2023-12-01
Fix type cast for normalizing alter subscription statement

anuraaga:
I am compiling this library to WebAssembly and ran into this issue. This test failed for me

https://github.com/pganalyze/libpg_query/blob/15-latest/test/normalize_tests.c#L30

The bug is obvious, though I'm not sure what it is about the WebAssembly toolchain that causes it to break vs how these tests are normally run, it could be the usage of musl as libc or just the LLVM Wasm backend itself.


lfittl:
Good catch!

I suspect this might be a struct alignment detail that behaves differently in the WASM toolchain, thus causing a problem only in that case. It was by sheer luck that this worked otherwise.

lfittl (approved):

PR #229, opened 2023-12-22
Additional deparser fixes for Postgres 16

lfittl:
On top of https://github.com/pganalyze/libpg_query/pull/225.

This makes all SQL contained in Postgres 16 regression tests work successfully with the deparser.


src/postgres_deparse.c
```diff
@@ -246,89 +251,144 @@ static void deparseAnyNameSkipLast(StringInfo str, List *parts)
 	}
 }
 
-// "a_expr" / "b_expr" in gram.y
-static void deparseExpr(StringInfo str, Node *node)
```
lfittl, line 250: This refactoring maps the deparser handling of a/b/c type expressions closer to how `gram.y` is written, and helps us only add parentheses around `a_expr` type expressions when needed (i.e. when they are used within the context of a `c_expr`).

msepga (approved):

PR #232, opened 2023-12-30
32-bit support

### LINKED ISSUE #231: 32-bit support
When building for Wasm in go-pgquery, I hadn't noticed the values in `pg_config.h` that are hard-coded for 64-bit architectures (Wasm is 32-bit).

https://github.com/pganalyze/libpg_query/blob/16-latest/src/postgres/include/pg_config.h#L680

With PG 15, this didn't seem to cause any issues, but when updating to PG 16 I found compilation errors relating to `sizeof(DATUM)` being the same as `sizeof(int)` which makes sense given that configuration. So I patched the config with various values for a 32-bit architecture

https://github.com/wasilibs/go-pgquery/blob/main/buildtools/wasm/patch.txt#L67

I was a bit surprised to see I had to implement `pg_popcount32` - though IIUC, the source import process effectively compiles the code, meaning that only code referenced from the config end up in this repo, so 32-bit codepaths get erased.

I was wondering if it wouldn't be possible to include at least `pg_popcount32` in the source? That allows only needing to do the config change for 32-bit, without actual business logic changes.

If it's not a priority to support 32-bit, no worries just wanted to check, and also leave a record for anyone else compiling to Wasm.

lfittl:
Fixes #231

### DIFF
--- a/protobuf/pg_query.proto
+++ b/protobuf/pg_query.proto
@@ -1272,7 +1272,7 @@ message RTEPermissionInfo
 {
   uint32 relid = 1 [json_name="relid"];
   bool inh = 2 [json_name="inh"];
-  int64 required_perms = 3 [json_name="requiredPerms"];
+  uint64 required_perms = 3 [json_name="requiredPerms"];
   uint32 check_as_user = 4 [json_name="checkAsUser"];
   repeated uint64 selected_cols = 5 [json_name="selectedCols"];
   repeated uint64 inserted_cols = 6 [json_name="insertedCols"];
--- a/Makefile
+++ b/Makefile
@@ -127,54 +127,11 @@ $(PGDIR):
 	cd $(PGDIR); make -C src/port pg_config_paths.h
 	cd $(PGDIR); make -C src/backend generated-headers
 	cd $(PGDIR); make -C src/backend parser-recursive # Triggers copying of includes to where they belong, as well as generating gram.c/scan.c
-	# This causes compatibility problems on some Linux distros, with "xlocale.h" not being available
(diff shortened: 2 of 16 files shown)


protobuf/pg_query.proto
```diff
@@ -1272,7 +1272,7 @@ message RTEPermissionInfo
 {
   uint32 relid = 1 [json_name="relid"];
   bool inh = 2 [json_name="inh"];
-  int64 required_perms = 3 [json_name="requiredPerms"];
+  uint64 required_perms = 3 [json_name="requiredPerms"];
```
lfittl, line 1275: As noted in the commit message, this is okay to do only because this struct isn't actually used in raw parse trees. Otherwise this would be an API break (which may be okay in the context of how we use these protobufs, but at least we should talk about doing this outside of major releases).

lelit:
FYI, I tried out this PR and it fixes the problem, all [CI tests](https://github.com/lelit/pglast/actions/runs/7363134571/job/20042296417#step:5:1136) are green again!
Thank you!

anuraaga (commented):
It's a bit tricky to verify this with go-pgquery since I use the C sources from pg_query_go instead of these ones to make sure they align, but I think this should fix my issue well. Thanks a lot!

msepga (approved):

anuraaga:
Just to close the loop, after updating I could remove my 32-bit support patch. Thanks!

https://github.com/wasilibs/go-pgquery/pull/22/files#diff-4d79e217dbbb7c7f611bfd2e2a6111d14e17bff3f3dbafac24d6314e5eeafc03L58

PR #233, opened 2024-01-01
Add support for running on Windows

### LINKED ISSUE #44: Windows Support
Issue to track support of Windows - currently it is not supported, in part due to build process, in part because the code extraction from Postgres intentionally ignores Windows code.

lfittl:
Fixes #44


Makefile.msvc
```diff
@@ -0,0 +1,108 @@
+# Makefile for "nmake", part of Microsoft Visual Studio Compiler (MSVC) on Windows
```
seanlinsley, line 1: I wonder if we should go ahead and switch to a platform-independent build script to avoid makefile drift. Since we already include Ruby, maybe it could be a set of rake tasks?

lfittl: Fair point re: Makefile drift. I wonder if there is a way to maintain the list of tests (and examples) another way, or just discover all of them in the test/examples directories, to avoid having to add new tests to both.

Generating Makefiles could work, but I worry that the indirection (i.e. having to go through a generation script first) isn't obvious - i.e. I'd prefer if we could find a way to avoid that extra step.

msepga: What about something like CMake / Meson, which generates `Makefile`s on *nix platforms and NMake specific `Makefile`s on Windows?

msepga (approved):
👍

Makefile.msvc
lfittl: Postgres itself started using Meson for this purpose, but that of course then introduces a build-time dependency (which I'm 50/50 on whether its okay to add).

I think for now I'll not adjust this since this is already a massive PR, but I think we could come back to this at a future time.

One approach to solving the duplication for tests/examples specifically could be to have a Makefile in the `tests` and `examples` folders that can run on either platform. The main motivation for not doing that in the root folder `Makefile` is that there are a bunch of features not supported by `nmake` that make it hard to write a portable Makefile that can do the code generation/etc conditionally.

PR #239, opened 2024-02-16
Fix: ar variable in Makefile (back patch to v13)

gregnr:
Applies the same patch as https://github.com/pganalyze/libpg_query/pull/217 but to v13 (in order to use in v13 version of [libpg-query-node](https://github.com/pyramation/libpg-query-node)).

Replaces the hardcoded `AR` variable:
```
AR = ar rs
```

with a reference to the upstream inherited `AR` variable:

```
AR ?= ar
AR := $(AR) rs
```

so that Emscripten's `emar` replacement is used.


lfittl (approved):
:+1: Thanks for the contribution!

Issue #243, opened 2024-04-16
Parser incorrectly parses queries with negative and zero values


MGorkov:
The following queries produces an invalid tree when run through the parser :
SELECT 0;
SELECT -1;

Parser output:
`
{
  version: 160001,
  stmts: [
    {
      stmt: {
        SelectStmt: {
          targetList: [
            {
              ResTarget: {
                val: { A_Const: { ival: {}, location: 7 } },
                location: 7
              }
            }
          ],
          limitOption: 'LIMIT_OPTION_DEFAULT',
          op: 'SETOP_NONE'
        }
      }
    }
  ]
}
`
In version 15 these queries also produce incorrect tree, but version 14 is correct.

Issue #160, opened 2022-11-10
plpgsql parsing error when defining alias for non-named parameter (e.g. _a alias for $1)


wesselvdv:
Trying to parse the following example gives the error `variable "$1" does not exist`

```plpgsql
CREATE OR REPLACE FUNCTION public.function_a(
    TEXT,
    TEXT
) RETURNS TEXT
    LANGUAGE plpgsql
AS
$$
DECLARE
    _a ALIAS FOR $1;
    _b ALIAS FOR $2;
BEGIN
    RETURN _b;
END;
$$;

SELECT
    public.function_a('aaa', 'bbb');
```

PR #244, opened 2024-04-16
fix some plpgsql errors

MGorkov:
This patch fixes the [#243](https://github.com/pganalyze/libpg_query/issues/243) , [#160](https://github.com/pganalyze/libpg_query/issues/160) , [#248](https://github.com/pganalyze/libpg_query/issues/248) issues and the following errors:
"variable "$1" does not exist"
"collations are not supported by type -"
"variable "ref" must be of type cursor or refcursor"
missing expression at or near ";"

changes in JSON:
1. add declared variable with "alliasfor" parameter pointing to associated variable, e.g. DECLARE _a ALIAS FOR $1;
```
    {
      PLpgSQL_var: {
        refname: '_a',
        aliasfor: '$1',
        lineno: 3,
        datatype: { PLpgSQL_type: { typname: 'text' } }
      }
```
2. add "retvarno" parameter for RETURN NEXT statement
```
    {
      PLpgSQL_stmt_return_next: { lineno: 6, retvarno: 2 }
    }
```


lfittl:
Thanks for the patch, and conceptually looks good!

However it appears there is a problem with the `protobuf-cpp` tests here, blocking a merge.

I believe this is actually a larger issue with how we handle `oneof` in our JSON output, I started fixing that in #246, pulling in your commit there as well. Assuming I can work out the open issues on that PR I'll close this here and fix it by merging the other PR.

lfittl:
@MGorkov Thanks for your contributions in this PR! I ended up reworking them a bit and merging them in #263.

I did not take all changes, but tried to keep it more narrow to the parts that looked most relevant to make parsing work in cases where it didn't work before.

If there are additional changes you'd like to see, could you propose these in a new PR? I'll close this here for now, since I believe the main issues are addressed.

PR #246, opened 2024-04-26
Avoid json outfuncs key duplication

lfittl:
Reviewing #244 I noticed that we appear to incorrectly wrap `ival`, `sval`, etc twice when outputting `A_Const` nodes. That does not match the Protobuf JSON approach, which intends to have a `oneof` represented the same way as having a simple field of that type (see https://github.com/protocolbuffers/protobuf/issues/3928#issuecomment-1505676321).

This change now matches the manual JSON output to the generate JSON protobuf. This changes
an output of `{"A_Const":{"ival":{"ival":123}}}` to be `{"A_Const":{"ival":123}}` instead.

This is a breaking change in the JSON output. As such, this would warrant a minor version bump (or we could hold off until the Postgres 17 release). Note that this does not change anything in the binary Protobuf output, which is used by most wrapper libraries.


lfittl:
Hmm, it looks like `protobuf-cpp` does not follow the spec here, since it does output the keys twice (see test failures). Need to debug this further.

msullivan:
We (https://github.com/edgedb/edgedb) use the JSON output and are trying to upgrade to 16 but are blocked by the int constant thing. We would definitely appreciate a fix on 16 (though we can make a temporary fork with a direct fix if needed)

msullivan:
It looks like the mismatch that was happening with protobuf wasn't actually the double key situation, but that protobuf still left out the value on `0`

msullivan:
I've submitted https://github.com/pganalyze/libpg_query/pull/249

lfittl:
Closing this for now since the integer constant issue was resolved separately, and the JSON duplication here seems best to keep here for now, to match `protobuf-cpp`.

PR #249, opened 2024-05-09
Fix parsing of negative integer constants

### LINKED ISSUE #243: Parser incorrectly parses queries with negative and zero values

msullivan:
Currently the value of negative constants are omitted. Only omit *0* constants.

An earlier attempt (#244) caused a mismatch with the protobuf json output, since that wants to omit values that are equal to the default value.

I don't *really* think that is particularly sensible behavior for an integer constant in an AST, but oh well.

Fixes #243.


aleclarson (approved):

lfittl (approved):
:+1: Thanks for figuring this out!

PR #255, opened 2024-08-06
Add function to normalize only utility statements

seanlinsley:
This PR introduces a `pg_query_normalize_utility` function which only normalizes utility statements. We're adding this so the pganalyze collector can redact database secrets from the Postgres logs in the case that `log_min_duration_statement` is set to zero.

### DIFF
--- a/pg_query.h
+++ b/pg_query.h
@@ -96,6 +96,7 @@ extern "C" {
 #endif
 
 PgQueryNormalizeResult pg_query_normalize(const char* input);
+PgQueryNormalizeResult pg_query_normalize_utility(const char* input);
 PgQueryScanResult pg_query_scan(const char* input);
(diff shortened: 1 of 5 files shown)


pg_query.h
```diff
@@ -96,6 +96,7 @@ extern "C" {
 #endif
 
 PgQueryNormalizeResult pg_query_normalize(const char* input);
+PgQueryNormalizeResult pg_query_normalize_utility(const char* input);
```
seanlinsley, line 99: If there's potential for more variants in the future, we should follow the `pg_query_parse_opts` pattern and use an enum. But unless we want to offer fine-grained normalization (e.g. normalizing specific statement types like `SelectStmt`), this probably isn't needed.

msakrejda (approved):

PR #256, opened 2024-08-17
Fix datatype handling for plpgsql function parameters

### LINKED ISSUE #126: Providing types for plpgsql function signature
`"UNKNOWN"` is currently returned for all types in a plpgsql function signature because of this:

https://github.com/pganalyze/libpg_query/blob/d1d0186b80c1fbf496265789af89da4e7ca890ab/src/pg_query_parse_plpgsql.c#L193

Which calls a mocked function:

https://github.com/pganalyze/libpg_query/blob/e9e4ba9ffcbac02bce6656ecabd22c2b54eca1b2/scripts/extract_source.rb#L470

We see that UNKNOWN is hardcoded because of `typ->typname = pstrdup("UNKNOWN");`.

It is quite easy to change plpgsql_build_datatype into something that returns a `PLpgSQL_type` where `typname` is populated with a given type. Say we have `create function test(x customtype, y int) ....`. For x the type will be customtype, while it will be pg_catalog.int4 for y because of the way the parser works. Question: Is that the reason this isn't implemented - that built in types are turned into something else for function signatures, while simply being provided as-is from the DECLARE section?

I can't figure out what ideally should be returned here. We do get the original location of the type in the function signature, and can use that to extract the actual written type, like int instead of pg_catalog.int4. It's not super straight forward with signatures like `create function complicated(x character(100), y time with time zone default '...', z "mySchema" . sometype[])` etc., but not knowing what would be the right thing to use here, I'd rather ask if some thought have gone into this, or what opinions you have?

svenklemm:
Currently all plpgql function parameters are returned as UNKNOWN. This patch changes function parameters to return actual type names used in the function definition.

Fixes #126


lfittl:
@svenklemm Thanks for the contribution - @msepga will take a look at this.

svenklemm:
@msepga any update?

svenklemm:
@lfittl looks like @msepga is MIA

lfittl:
> @lfittl looks like @msepga is MIA

Just got a lot on our plate, but I'll make sure we can get this reviewed in the next days. Thanks for your patience!

msepga (commented):
Thanks for the contribution and patience, @svenklemm! The mock LGTM :+1:

One note, looks like running `make extract_source` seemed to drop `quote_qualified_identifier` in the source here. I've added a call to resolve the symbol:

```diff
 # PL/pgSQL Parsing
 runner.deep_resolve('plpgsql_compile_inline')
 runner.deep_resolve('plpgsql_free_function_memory')
+runner.deep_resolve('quote_qualified_identifier')
```

I'll approve this PR once the tests pass with the new commit.

For reference, here's the original diff I get after running `make extract_source`, without the new commit:

<details>


```diff
diff --git a/src/postgres/src_backend_utils_adt_ruleutils.c b/src/postgres/src_backend_utils_adt_ruleutils.c
index 30df7c3..9257991 100644
--- a/src/postgres/src_backend_utils_adt_ruleutils.c
+++ b/src/postgres/src_backend_utils_adt_ruleutils.c
@@ -1718,18 +1718,7 @@ quote_identifier(const char *ident)
  * Return a name of the form qualifier.ident, or just ident if qualifier
  * is NULL, quoting each component if necessary.  The result is palloc'd.
  */
-char *
-quote_qualified_identifier(const char *qualifier,
-						   const char *ident)
-{
-	StringInfoData buf;
 
-	initStringInfo(&buf);
-	if (qualifier)
-		appendStringInfo(&buf, "%s.", quote_identifier(qualifier));
-	appendStringInfoString(&buf, quote_identifier(ident));
-	return buf.data;
-}
 
 /*
  * get_relation_name
diff --git a/src/postgres/src_pl_plpgsql_src_pl_comp.c b/src/postgres/src_pl_plpgsql_src_pl_comp.c
index 197de46..f99fab0 100644
--- a/src/postgres/src_pl_plpgsql_src_pl_comp.c
+++ b/src/postgres/src_pl_plpgsql_src_pl_comp.c
@@ -879,6 +879,7 @@ plpgsql_build_recfield(PLpgSQL_rec *rec, const char *fldname)
  * It can be NULL if the type could not be a composite type, or if it was
  * identified by OID to begin with (e.g., it's a function argument type).
  */
+
 PLpgSQL_type * plpgsql_build_datatype(Oid typeOid, int32 typmod, Oid collation, TypeName *origtypname)
 {
 	PLpgSQL_type *typ;
@@ -923,6 +924,8 @@ PLpgSQL_type * plpgsql_build_datatype(Oid typeOid, int32 typmod, Oid collation,
 	return typ;
 }
 
+
+
 /*
  * Utility subroutine to make a PLpgSQL_type struct given a pg_type entry
  * and additional details (see comments for plpgsql_build_datatype).
```

</details>

msepga (approved):

PR #259, opened 2024-09-16
add pg_query.rs to README

lemonadern:
added a mention of `pg_query.rs` to the README

https://github.com/pganalyze/pg_query.rs


lfittl:
Good catch, thanks for noticing, and the PR!

PR #267, opened 2024-10-20
Fix cursor quoting in FetchStmt

rjuju:
Regression tests updated to test all cursor related statement with a name that requires quoting.


rjuju:
Note that the problem exist in all version, so it should be applied to all the branches.

msepga:
Thanks for the change, @rjuju! I've rebased this and moved the test case to `deparse_tests.c`, as the `test/sql/` folder is extracted directly from the Postgres source tree.

msepga (approved):

PR #269, opened 2024-12-15
Fix: deparse DEFAULT in ColumnDef `AT TIME ZONE`

tylergannon:
I submitted a bug report earlier today in the [pg_query_go issues](https://github.com/pganalyze/pg_query_go/issues/126), then decided to have a look and see if I could fix it.

This fixes a bug in which DEFAULT expressions are deparsed into invalid SQL.

I added test cases to `deparse_tests.c`, which are corrected in the suggested fix.

I'm not sure about whether my manner of detecting the situation will suit you, but I hope we can get something going here in the way of a fix.  I'm trying to release some software that depends on being able to deparse queries of this shape, and am going to have to, for the time being, devise some kind of workaround in order to ship.

Thanks!  Let me know if I can help in any way.

### DIFF
--- a/src/postgres_deparse.c
+++ b/src/postgres_deparse.c
@@ -2703,6 +2703,30 @@ static void deparseFuncCall(StringInfo str, FuncCall *func_call)
 		 */
 		Expr* e;
 		bool isLocal = list_length(func_call->args) == 1;
+		/* 
+		 * time values expressed `AT TIME ZONE` inside a DEFAULT value expression must be surrounded by parentheses
+		 * in order for the deparsed DEFAULT to be valid syntax.
+		 * The following is a simple lookback into the deparsed SQL, to see if is expression immediately follows
+		 * a DEFAULT declaration.
+		 */
+		bool isDefaultValue = false;
+
+		{
+			int end = str->len;
+			// Step back over trailing spaces
+			while (end > 0 && isspace((unsigned char) str->data[end - 1]))
+				end--;
+
+			// Now check if, just before these trailing spaces, we have "DEFAULT"
+			if (end >= 7 && strncmp(str->data + (end - 7), "DEFAULT", 7) == 0)
+			{
+				isDefaultValue = true;
+			}
+		}
+
+		if (isDefaultValue) {
+			appendStringInfoChar(str, '(');
+		}
 
 		if (isLocal)
 			e = linitial(func_call->args);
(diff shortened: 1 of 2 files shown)


lelit:
I did not check your implementation, but AFAICT also the following case requires parens:
```sql
create table my_table (created_at timestamptz not null default ('1 hour'::interval + current_timestamp at time zone 'UTC'))
```
so I'm not sure the strategy to look for an _immediately preceeding_ `DEFAULT` is correct.

tylergannon:
@lelit good point, thanks for responding.  I added that test case and of course it isn't repaired by my attempted fix.  What would you suggest?

If it was totally up to me, I'd just parenthesize this type of construct in all cases, in order to ensure that deparse always returns valid SQL, but that is bound to throw off a lot of tests downstream.

Meanwhile, I don't know of there being any contextual info on the StringInfo or FuncCall objects, that would reveal that this is happening inside of a DEFAULT declaration.

Is there any other category I could ascribe to this positioning where the parentheses are required around the expression, which I could use to detect it?

lelit:
I'm afraid I don't have a quick answer, because I don't know very well the internals of the deparsing tool: as you say there is no (AFAIK) any generic way to pass down what you call _contextual info_ to lower nodes, and that complicates the solution.
One example of a similar problem is in the function [deparseTypeCast()](https://github.com/pganalyze/libpg_query/blob/17-latest/src/postgres_deparse.c#L3690), that accepts a _context_ from higher levels to differentiate its output, but in this case it is tricky, because that context (computed in the `deparseConstraint()` function) should be passed down to a possibly wide range of nodes that may wrap the actual _leaf_ `deparseFuncCall()` function...
Sorry for not being of a great help, maybe Lukas or msepga may say something more interesting...

tylergannon:
I appreciate the help, and I'd rather get the solution right then to persist with a poor workaround.  I've got my own application patched with that solution and it works for me FTTB so I've no pressure to solve it immediately and can help to solve it _right_ if I can.

I'll check out the deparseTypeCast function and its call sites, and see if I can propose something based on what I learn there.

lfittl:
@tylergannon Thanks for your contribution! I agree with @lelit that we need to think about the other cases as well, and making use of context is a good idea.

Since this is a bit hard to work on I believe for someone new to the deparser code base, I took a stab at this in #270 and did some related refactoring as well. Let me know if this looks good to you / any further cases you think that need to be tested.

tylergannon:
superseded by #270

PR #270, opened 2024-12-17
Improve deparser expression handling, fix deparsing of DEFAULT expressions classified as "a_expr"

lfittl:
Alternate to #269 that fixes the discussed edge cases there as well.

Best reviewed by looking at individual commits.


test/deparse_tests.c
```diff
@@ -409,7 +409,11 @@ const char* tests[] = {
   "CREATE INDEX myindex ON public.mytable USING btree (col1, (col2::varchar) varchar_pattern_ops)",
   "SELECT * FROM CAST(1 AS text)",
   "DECLARE \"Foo1\" SCROLL CURSOR FOR SELECT * FROM tenk1 ORDER BY unique2",
-  "FETCH BACKWARD 23 \"Foo1\""
+  "FETCH BACKWARD 23 \"Foo1\"",
+  "CREATE TABLE my_table (created_at timestamptz NOT NULL DEFAULT (now() AT TIME ZONE 'UTC'))",
+  "CREATE TABLE my_table (created_at timestamptz NOT NULL DEFAULT (now() AT LOCAL))",
+  "CREATE TABLE my_table (created_at timestamptz NOT NULL DEFAULT '1 hour'::interval + (current_timestamp AT TIME ZONE 'UTC'))",
```
tylergannon, line 415:
```suggestion
  "CREATE TABLE my_table (created_at timestamptz NOT NULL DEFAULT '1 hour'::interval + (current_timestamp AT TIME ZONE 'UTC'))",
  "ALTER TABLE my_table ADD COLUMN created_at timestamptz NOT NULL DEFAULT '1 hour'::interval + (current_timestamp AT TIME ZONE 'UTC')",
```

Just to make explicit that the fix includes ADD COLUMN with DEFAULT

tylergannon (changes requested):
Thanks for this!  Enjoyed reading through the commits and it helped me get a better sense of the flow through the deparser code.  True, as a first-time committer I would not have landed a correct solution like this.

Not sure if it adds any information to your tests, but I added one example to the deparser test to explicitly support ALTER TABLE ... ADD COLUMN.

msepga (approved):

test/deparse_tests.c
lfittl: Yep, good idea!

lfittl:
@tylergannon Thanks again for raising this - merged!

FYI, we're currently planning to tag a libpg_query version with this (that also brings it to pg_query_go) sometime in the early new year, since there are a few other minor things we may take care of before the next release.

My understanding is you have a working solution for your project in the interim, but let us know if it'd be important for you to get an earlier patch release.

PR #274, opened 2025-02-27
Add `pg_query_parse_summary()` and supporting code

duckinator:
NOTE: Any code related to `truncated_query` has been commented out in a way where `git grep 'UNIMPLEMENTED(truncated_query)'` should find it. So the code that was already in place is kept, but it's not in any public APIs. It's not ideal, but it's the cleanest way I could think of doing it.

Remaining tasks:
- [x] collecting "statement type" information in C + passing to Rust via protobuf
- [x] finish the big switch/case statement (see the other list below)
- [x] processing `subselect_items` (or equivalent code)
- [x] processing CTEs
- [x] processing `filter_columns`
- [x] processing `aliases`
- [x] Enable all existing Rust tests for parse summaries
- [x] import `makeRangeVarFromNameList()` _properly_
- [x] tests in C
- [x] reformat code so `{` is put on its own line, to match Postgres style
- [x] use valgrind and such to check for memory leaks
- [x] pganalyze/pg_query.rs#65

Branches of the big switch/case statement that have been ported:

- [x] `List`
- [x] `SelectStmt`
- [x] `InsertStmt`
- [x] `UpdateStmt`
- [x] `DeleteStmt`
- [x] `CopyStmt`
- [x] `AlterTableStmt`
- [x] `CreateStmt`
- [x] `CreateTableAsStmt`
- [x] `TruncateStmt`
- [x] `ViewStmt`
- [x] `IndexStmt`
- [x] `CreateTrigStmt`
- [x] `RuleStmt`
- [x] `VacuumStmt`
- [x] `RefreshMatViewStmt`
- [x] `DropStmt`
- [x] `GrantStmt`
- [x] `LockStmt`
- [x] `ExplainStmt`
- [x] `CreateFunctionStmt`
- [x] `RenameStmt`
- [x] `PrepareStmt`
- [x] `RawStmt`
- [x] `ColumnRef`
- [x] default


scripts/generate_protobuf_and_funcs.rb
```diff
@@ -441,6 +441,39 @@ def generate!
   // Named tokens in scan.l
   #{@scan_protobuf_tokens.join("\n  ")}
 }
+
+message SummaryResult {
+  message SummaryResultTable {
+    string table = 1;
+    // FIXME(duckinator): I AM NOT CONFIDENT Node IS RIGHT. THIS NEEDS TO BE VERIFIED.
+    Node context = 2;
```
duckinator: It is extremely unclear to me if `Node` is correct here, but it's the closest Protobuf thing I could find to the `Context` enum in Rust:

https://github.com/pganalyze/pg_query.rs/blob/23ab4ae1c989fe0dea919d4fee043993e97f6223/src/node_enum.rs#L5-L12

Is this right, or does there need to be a new thing added that's equivalent to `Context`?

scripts/generate_protobuf_and_funcs.rb
```diff
@@ -441,6 +441,39 @@ def generate!
...
+  map<string, string> aliases = 2;
+
+  // cte_names: Vec<String>
+  repeated string cte_names = 3;
+
+  // functions: Vec<(String, Context)>
+  message SummaryResultFunction {
+    string name = 1;
+    // FIXME(duckinator): Just like with SummaryResultTable, I am *NOT* confident using Node is right, and this needs to be verified.
+    Node context = 2;
```
duckinator: Same question here about the protobuf equivalent of `Context`. Is that `Node`, or do I need to add something else?

scripts/generate_protobuf_and_funcs.rb
```diff
@@ -441,6 +441,39 @@ def generate!
...
+  }
+  SummaryResultFunction functions = 4;
+
+  // filter_columns: HashSet<(Option<String>, String)>
+  message SummaryResultFilterColumn {
+    // FIXME(duckinator): What should a and b be named?
+    // In the Rust version, they *aren't* named -- it's a Set of Tuples.
+    optional string a = 1;
+    string b = 2;
```
duckinator: In Rust, this is a `filter_columns: HashSet<(Option<String>, String)>`.

Notably, the inner parts are in a tuple so it's not named. I couldn't determine [from the context in pg_query.rs](https://github.com/pganalyze/pg_query.rs/blob/23ab4ae1c989fe0dea919d4fee043993e97f6223/src/parse_result.rs#L150-L161) what would make sense as a name.

Is `a` and `b` fine, or do you have better suggestions?

scripts/generate_protobuf_and_funcs.rb
```diff
@@ -441,6 +441,39 @@ def generate!
   // Named tokens in scan.l
   #{@scan_protobuf_tokens.join("\n  ")}
 }
+
+message SummaryResult {
+  message SummaryResultTable {
```
duckinator: In general, do y'all have a preference for nested message types, or non-nested?

If you do have a preference, is it just a style thing, or is there a notable difference (e.g. in performance or something)?

scripts/generate_protobuf_and_funcs.rb
seanlinsley: They're table name and column name:
https://github.com/pganalyze/pg_query.rs/blob/23ab4ae1c989fe0dea919d4fee043993e97f6223/tests/filter_column_tests.rs#L10-L21

Which could either be `table` and `column`, or `table_name` and `column_name`

scripts/generate_protobuf_and_funcs.rb
seanlinsley: I don't believe there's a performance difference, but one benefit to nesting messages is you can use a less-specific message name, e.g. renaming `SummaryResultTable` to `Table`, without running into naming conflicts from other messages.

scripts/generate_protobuf_and_funcs.rb
seanlinsley: `Node` represents the full parse tree, so wouldn't work here. Instead you should add a protobuf enum that matches the `Context` enum in Rust.

pg_query.h
```diff
@@ -119,6 +125,8 @@ PgQuerySplitResult pg_query_split_with_parser(const char *input);
 
 PgQueryDeparseResult pg_query_deparse_protobuf(PgQueryProtobuf parse_tree);
 
+PgQuerySummaryParseResult pg_query_parse_summary(const char* input, int parser_options, int truncate_limit);
```
seanlinsley: We might want to add a `_protobuf` suffix on the function name to follow the existing naming convention. Thoughts @lfittl?

Also `parse_summary` might be too verbose. Could it just be `summary`?

Then we'd have `pg_query_summary_protobuf`

duckinator: I was wondering about removing the `parse_` part of the name as well. I think `pg_query_summary_protobuf` would be fine, but if @lfittl has other options I'm open to them. :+1:

lfittl: As discussed we can leave this as `pg_query_parse_summary` since that matches how we named `pg_query_scan` (which also returns a Protobuf) -- the `_protobuf` suffix on parse is mostly for historic reasons.

Also as discussed we're keeping "parse" in the function name since the main function here is still parsing (vs scanning/normalizing).

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+  enum Context {
+    None = 0;
+    Select = 1;
+    DML = 2;
+    DDL = 3;
+    Call = 4;
+  }
+
+  message Table {
+    string table = 1;
```
lfittl: As discussed, probably best to follow the Ruby structure here, which has a few more fields:

https://github.com/pganalyze/pg_query/blob/main/lib/pg_query/parse.rb#L319

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+    Context context = 2;
+  }
+  // tables: Vec<(String, Context)>
+  repeated Table tables = 1;
+
+  // aliases: HashMap<String, String>
+  map<string, string> aliases = 2;
```
lfittl, line 4137: For clarity, the alias value here is the table name (i.e. `schema.table` or just `table` depending on how its written in the query).

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+
+  // cte_names: Vec<String>
+  repeated string cte_names = 3;
+
+  // functions: Vec<(String, Context)>
+  message Function {
+    string name = 1;
```
lfittl, line 4142: As discussed, would be good to also add `function_name` and `schema_name` here, just to make it easier for the caller to work with the result (and avoid an unnecessary split on the qualified name).

To get those values:
- `function_name` is the last entry in the `funcname` list
- `schema_name` is the second to last entry in the `funcname` list (and may not be present)

(there could also be a third entry in the beginning, which is the catalog or database name, but we can ignore that)

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+    Context context = 2;
+  }
+  Function functions = 4;
+
+  // filter_columns: HashSet<(Option<String>, String)>
+  message FilterColumn {
+    // protobuf-c doesn't support optional fields, so I guess just... make
+    // table an empty string if it should be None-equivalent?
+    // Upstream issue: https://github.com/protobuf-c/protobuf-c/issues/476
```
lfittl: I think using empty strings here is the best solution - we'll have to do the same for some other fields (e.g. `Table.schema_name`)

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+    string table = 1;
+    string column = 2;
```
lfittl:
```suggestion
    string schema = 1; /* optional, may be empty */
    string table = 2; /* optional, may be empty */
    string column = 3;
```

Also we should check if the table name is an alias, and if it is look it up and populate `table` as well as `schema` (if applicable).

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,44 @@ enum Token {
...
+  }
+  repeated FilterColumn filter_columns = 5;
+}
```
lfittl, line 4158:
```suggestion
  string truncated_query = 6; /* optional, only returned if truncation limit is not -1, otherwise empty */
}
```

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,50 @@ enum Token {
...
+    string name = 1;
+    string function_name = 2;
+    string schema_name = 3; // optional
+    Context context = 4;
+  }
+  repeated Function functions = 4;
+
+  message FilterColumn {
+    string schema = 1; // optional
+    string table = 2; // optional
```
lfittl:
```suggestion
    string schema_name = 1; // optional
    string table_name = 2; // optional
```

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,50 @@ enum Token {
...
+    Select = 1;
+    DML = 2;
+    DDL = 3;
+    Call = 4;
+  }
+
+  message Table {
+    string name = 1;
+    string schema_name = 2;
+    string rel_name = 3;
```
lfittl:
```suggestion
    string name = 1;
    string schema_name = 2;
    string table_name = 3;
```

protobuf/pg_query.proto
```diff
@@ -4108,3 +4108,50 @@ enum Token {
...
+  }
+  repeated Function functions = 4;
+
+  message FilterColumn {
+    string schema = 1; // optional
+    string table = 2; // optional
+    string column = 3;
+  }
+  repeated FilterColumn filter_columns = 5;
+  string truncated_query = 6; /* optional, empty if truncation limit is -1 */
```
lfittl: Whilst we'll work on this next, it might be best to remove this from the Protobuf for now so its not confusing anyone who doesn't know about it being a stub.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
+#include "pg_query.h"
```
lfittl: As discussed we can rename `pg_query_parser_summary` to `pg_query_summary` (and the test from `parse_summary.c` to `summary.c`), that'll make the naming more clear when we have more related, yet separate, unit files (like `pg_query_summary_statement_types.c`).

pg_query.h
lfittl: This naming discussion came up again in the context of having multiple unit files, like we now have `pg_query_summary_statement_types.c`. Because of that, I think it would actually make sense to shorten this to `pg_query_summary` after all.

Whilst past-Lukas had a point that this is specific to parsing, I don't actually think we'll add a scan summary or normalize summary ever, so it seems fine to just call this "summary", even though technically its a summary of the parse result.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+#include "protobuf/pg_query.pb-c.h"
+
+#include <unistd.h>
+#include <fcntl.h>
+#include <string.h>
+
+#include "pg_query_summary.h"
+
+// NOTE regarding cte_names and from_clause_items:
```
lfittl: As discussed it may make sense to move this closer to the place where the comment applies (e.g. before the second tree walk).

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+// 1. We process cte_names as we iterate through the tree.
+// 2. We only add items to tables if they are *not* in cte_names.
+// 3. CTEs can be referenced *after* table names.
+// 4. Because of this, we have to store from_clause_items and process it after the initial tree walk.
+//
+// If we try to do it entirely on the initial pass, we wind up CTEs being
```
lfittl:
```suggestion
// If we try to do it entirely on the initial pass, we wind up with CTEs being
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+// included in state->tables, and thus in summary->tables.
+
+// We need to declare this early because it gets called before it's defined.
```
lfittl: It should be fine to not have this comment.

However, we probably should add `static` keywords to any functions not actually used outside of this file.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+void add_table(WalkState* state, enum ContextType context, char* schemaname, char* relname);
+
+// small utility function for creating error messages.
```
lfittl: We could probably also not have this comment :)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+PgQueryError* pg_query_parse_summary_mkerror(const char* message)
```
lfittl:
```suggestion
PgQueryError* make_pg_query_error(const char* message)
```

lfittl: (we don't need the `pg_query_parse_summary` suffix for static functions that are local to this file)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+	return error;
+}
+
+// Given a Node (node) and WalkState (state), wraps node in a NodeWithContext
+// and adds it to the state->from_clause_items list.
+void add_from_clause_item(Node *node, enum ContextType context, WalkState *state) {
+	if (!node)
+		return;
+
+	NodeWithContext* nwc = palloc(sizeof(NodeWithContext));
```
lfittl: It looks like we don't set `type` here, and we're not using `palloc0`, so `type` is undefined (and may have arbitrary values). Are we actually using `type` anywhere?

src/pg_query_summary.h
```diff
@@ -0,0 +1,89 @@
...
+	List* subselect_items;
+	List* from_clause_items;
+	List* cte_names; // list of CommonTableExpr
+	List* filter_columns; // List of FilterColumn*
+	bool has_filter_columns;
+} WalkState;
+
+typedef struct {
+	// Include `NodeTag type` so this wrapper is as transparent as possible.
+	NodeTag type;
```
lfittl: Since this isn't actually used anymore, maybe its okay to not have this be transparent and instead require the caller to dig into `node` to be able to switch on the node type.

src/pg_query_parse_summary.c
lfittl: Per our discussion it looks like `type` isn't used, if we remove it from `NodeWithContext` this comment here is moot.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+}
+
+// Given a List*-of-Nodes (clause) and WalkState (state), wrap them all
+// in NodeWithContext structs and add them to the state->from_clause_items list.
+void add_from_clause_items(List *clause, enum ContextType context, WalkState *state) {
+	if (!clause)
+		return;
+
+	ListCell *lc = NULL;
+	foreach(lc, clause) {
```
lfittl: This is a stylistic detail but to be consistent with how Postgres source itself is formatted it would be good if we can put all `{` on their own line (i.e. both here, in function definitions, if conditions with multiple lines, etc)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+	NodeWithContext* nwc = palloc(sizeof(NodeWithContext));
+	nwc->node = node;
+	nwc->context = context;
+
+	state->from_clause_items = lappend(state->from_clause_items, nwc);
+}
+
+// Given a List*-of-Nodes (clause) and WalkState (state), wrap them all
+// in NodeWithContext structs and add them to the state->from_clause_items list.
+void add_from_clause_items(List *clause, enum ContextType context, WalkState *state) {
```
lfittl: Per discussion, we can use a typedef here to avoid the recurring `enum` specifier.

src/pg_query_summary.h
```diff
@@ -0,0 +1,89 @@
...
+	char *key;
+	char *value;
+} SummaryAlias;
+
+typedef struct {
+	char *name;
+	char *function_name;
+	char *schema_name; // optional
+	enum ContextType context;
+} Function;
```
lfittl: Maybe helpful to prefix this and `Table` with `Summary` as well, to avoid any future conflicts.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+	initStringInfo(&buf);
+
+	bool first = true;
+	ListCell* lc = NULL;
+	foreach(lc, funcname) {
+		// Add a dot separator before everything but the first item.
+		if (first)
+			first = false;
+		else
+			appendStringInfoChar(&buf, '.');
```
lfittl:
```suggestion
		if (list_cell_number(funcname, lc) != 0)
			appendStringInfoChar(&buf, '.');
```

(and we can probably skip the comment since the code is self-explanatory.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+		String* str = (String*)lfirst(lc);
```
lfittl:
```suggestion
		String* str = lfirst_node(String, lc);
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+		appendStringInfoString(&buf, quote_identifier(str->sval));
+	}
+
+	char* name = buf.data;
+	String* str_name = makeString(name);
+	Function* func = palloc(sizeof(Function));
+
+	func->name = str_name->sval;
```
lfittl:
```suggestion
	func->name = buf.data;
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+			first = false;
+		else
+			appendStringInfoChar(&buf, '.');
+
+		String* str = (String*)lfirst(lc);
+		appendStringInfoString(&buf, quote_identifier(str->sval));
+	}
+
+	char* name = buf.data;
+	String* str_name = makeString(name);
```
lfittl: These two can be dropped.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+	if (list_length(funcname) > 1) {
+		String* first = linitial(funcname);
+		func->schema_name = first->sval;
+	} else {
+		func->schema_name = NULL;
+	}
+
+	// TODO: Verify this definition of function_name is correct.
+	String* last = llast(funcname);
+	func->function_name = last->sval;
```
lfittl: This is correct, but I'd keep it shorter without the extra variable:

```suggestion
	func->function_name = llast(funcname)->sval;
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+	func->name = str_name->sval;
+
+	// TODO: Verify this definition of schema_name is correct
+	if (list_length(funcname) > 1) {
+		String* first = linitial(funcname);
+		func->schema_name = first->sval;
+	} else {
+		func->schema_name = NULL;
+	}
```
lfittl:
```suggestion
        if (list_length(funcname) == 3)
        	func->schema_name = lsecond(funcname)->sval;
        else if (list_length(funcname) == 2)
        	func->schema_name = linitial(funcname)->sval;
        else
          	func->schema_name = NULL;
```

(as discussed the three item case can happen when the database name is added in front)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+		}
+
+		case T_RenameStmt: {
+			RenameStmt* stmt = (RenameStmt*)node;
+			if (stmt->renameType == OBJECT_FUNCTION) {
+				ObjectWithArgs* obj = (ObjectWithArgs*)stmt->object;
+				// Add the original name.
+				add_function(obj->objname, ContextDDL, state);
+				// Add the new name.
+				add_function_cstr(stmt->newname, ContextDDL, state);
```
lfittl:
```suggestion
				add_function(list_make1(makeString(stmt->newname)), ContextDDL, state);
```

(that way we don't need the extra `add_function_cstr` helper)

src/pg_query_summary.h
```diff
@@ -0,0 +1,89 @@
...
+#include "pg_query_internal.h"
+#include <unistd.h>
+
+enum ContextType {
+	ContextNone = 0,
+	ContextSelect = 1,
+	ContextDML = 2,
+	ContextDDL = 3,
+	ContextCall = 4,
+};
```
lfittl:
```suggestion
typedef enum ContextType {
	CONTEXT_NONE = 0,
	CONTEXT_SELECT = 1,
	CONTEXT_DML = 2,
	CONTEXT_DDL = 3,
	CONTEXT_CALL = 4,
} ContextType;
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+	func->name = name;
+	func->schema_name = NULL;
+	func->function_name = NULL;
+	func->context = context;
+
+	state->functions = lappend(state->functions, func);
+}
+
+// This should eventually be equivalent to https://github.com/pganalyze/pg_query/blob/main/lib/pg_query/parse.rb#L98-L340
```
lfittl: I think we can drop this comment so it doesn't become stale.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+	// NOTE: If `raw_expression_tree_walker()` doesn't handle a particular
+	// node type, you will need to `return false` early to avoid passing it
+	// to the tree walker.
+	// If you do this, be sure to update `pg_query_summary_stmt_walk_impl()`.
+
+	// NOTE: The Ruby implementation manually handles being given a List.
+	//
+	// In this version, the raw_expression_tree_walker() call at the
+	// end of this function will walk any items in a T_List node for us.
```
lfittl: Whilst that makes sense I think we can drop this comment since its typical for Postgres treewalker use to rely on the treewalker to descend through standard types like `List`.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+	switch (nodeTag(node)) {
+		case T_CommonTableExpr: {
```
lfittl: Postgres source would put the `{` on the next line and indent it. Also, if there are no new variables defined inside the case statement, the `{` is omitted.

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+			CommonTableExpr* cte = (CommonTableExpr*)node;
+			state->cte_names = lappend(state->cte_names, strdup(cte->ctename));
+			break;
+		}
+
+		case T_SelectStmt: {
+			printf("found select statement\n");
```
lfittl: I'd remove all printf in the committed version :-)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+	// If you do this, be sure to update `pg_query_summary_stmt_walk_impl()`.
+
+	// NOTE: The Ruby implementation manually handles being given a List.
+	//
+	// In this version, the raw_expression_tree_walker() call at the
+	// end of this function will walk any items in a T_List node for us.
+
+	switch (nodeTag(node)) {
+		case T_CommonTableExpr: {
+			CommonTableExpr* cte = (CommonTableExpr*)node;
```
lfittl:
```suggestion
			CommonTableExpr* cte = (CommonTableExpr*) node;
```

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+			state->cte_names = lappend(state->cte_names, strdup(cte->ctename));
```
lfittl:
```suggestion
			state->cte_names = lappend(state->cte_names, pstrdup(cte->ctename));
```

(generally, we should use `pstrdup` to utilize Postgres arena memory allocation, unless the result is returned to the caller, which should not be the case here since we encode all of this in a protobuf struct and intend to free this memory before returning the Protobuf to the caller)

src/pg_query_parse_summary.c
```diff
@@ -0,0 +1,832 @@
...
+
+			if (stmt->op == SETOP_NONE)
+				add_from_clause_items(stmt->fromClause, ContextSelect, state);
+
+			// For SETOP_UNION, SETOP_EXCEPT, and SETOP_INTERSECT,
+			// the raw_expression_tree_walker() call at the end of
+			// this function will handle stmt->larg/stmt->rarg.
+
+			if (stmt->intoClause)
+				add_from_clause_item((Node*)stmt->intoClause->rel, ContextDDL, state);
```
lfittl:
```suggestion
				add_from_clause_item((Node*) stmt->intoClause->rel, ContextDDL, state);
```

src/pg_query_summary.h
```diff
@@ -0,0 +1,89 @@
...
+
+typedef struct {
+	List* tables; // List of Table*
+	List *aliases; // List of SummaryAlias*
+	List* functions; // List of Function*
+	List* subselect_items;
+	List* from_clause_items;
+	List* cte_names; // list of CommonTableExpr
+	List* filter_columns; // List of FilterColumn*
+	bool has_filter_columns;
```
lfittl: Maybe `save_filter_columns` to explain the behavior that happens when this is set.

protobuf/pg_query.proto
duckinator: Done, but see the note in the main PR description about this.

duckinator:
Everything from the initial (partial) code review should be resolved, aside from the part about putting `{` on its own line. I moved that to the TODO list in top comment so it doesn't get forgotten.

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+#include <string.h>
+
+#include "pg_query_summary.h"
+
+static void add_table(WalkState* state, ContextType context, char* schemaname, char* relname);
+
+static PgQueryError* make_pg_query_error(const char* message)
+{
+	PgQueryError* error = palloc(sizeof(PgQueryError));
+	error->message = pstrdup(message);
```
lfittl:
```suggestion
        // These are returned to the caller, so allocate them outside of the Postgres memory context
	PgQueryError* error = malloc(sizeof(PgQueryError));
	error->message = strdup(message);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+	return error;
+}
+
+// Given a Node (node) and WalkState (state), wraps node in a NodeWithContext
+// and adds it to the state->from_clause_items list.
+static void add_from_clause_item(void *node, ContextType context, WalkState *state) {
```
lfittl: As discussed, whilst it requires more casts, it would be more consistent with Postgres style to use the more specific pointer type here (`Node *`).

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		case T_CommonTableExpr: {
+			CommonTableExpr* cte = castNode(CommonTableExpr, node);
+			state->cte_names = lappend(state->cte_names, pstrdup(cte->ctename));
+			break;
+		}
+
+		case T_SelectStmt: {
+			SelectStmt *stmt = castNode(SelectStmt, node);
+
+			pg_query_summary_walk_impl((Node*) stmt->targetList, state);
```
lfittl: This shouldn't be necessary, since the `raw_expression_tree_walker` already takes care of this: https://github.com/postgres/postgres/blob/88f0fdabead51539aa5bdefcf188f07d4651ee10/src/backend/nodes/nodeFuncs.c#L4348C20-L4348C31

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+	// If you do this, be sure to update `pg_query_summary_stmt_walk_impl()`.
+
+	switch (nodeTag(node)) {
+		case T_CommonTableExpr: {
+			CommonTableExpr* cte = castNode(CommonTableExpr, node);
+			state->cte_names = lappend(state->cte_names, pstrdup(cte->ctename));
+			break;
+		}
+
+		case T_SelectStmt: {
```
lfittl:
```suggestion
		case T_SelectStmt: {
		        // Don't descend into sub-SELECTs in the WHERE clause when saving filter columns
			if (state->save_filter_columns)
				return true;
```

lfittl: Query we can test this with:

```
SELECT * FROM test WHERE col1 = (SELECT col2 FROM test2 WHERE col3 = 123)
```

Should return `filter_columns = ["col1", "col3"]` (but not col2)

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+			if (stmt->intoClause)
+				add_from_clause_item(stmt->intoClause->rel, CONTEXT_DDL, state);
+
+			if (stmt->whereClause) {
+				// If we have a WHERE clause, that means we have to check for filter columns.
+				//
+				// NOTE: the WHERE clause is walked twice, unfortunately.
+				//  I don't know how to avoid this. It doesn't *appear*
+				//  to cause problems, but it means there's definitely an
+				//  optimization opportunity available.
```
lfittl:
```suggestion
				// NOTE: the WHERE clause is intentionally walked twice, but when saving
				// filter columns we don't descend into sub-SELECTs
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		case T_DeleteStmt: {
+			DeleteStmt *stmt = castNode(DeleteStmt, node);
+			add_from_clause_item(stmt->relation, CONTEXT_DML, state);
+			add_from_clause_items(stmt->usingClause, CONTEXT_SELECT, state);
+			break;
+		}
+
+		case T_CopyStmt: {
+			CopyStmt* stmt = castNode(CopyStmt, node);
+			add_from_clause_item(stmt->relation, CONTEXT_SELECT, state);
```
lfittl:
```suggestion
			add_from_clause_item(stmt->relation, CONTEXT_SELECT, state);
			pg_query_summary_walk_impl((Node*) stmt->query, state);
```

Like in https://github.com/pganalyze/pg_query/blob/d8a6b8f2028978c3e9295b4129060fd808bd6989/lib/pg_query/parse.rb#L174C37-L174C52 we need to handle `COPY (SELECT * FROM tbl) TO file.test` type statements.

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+			add_from_clause_item(castNode(CreateStmt, node)->relation, CONTEXT_DDL, state);
+			return false; // Not handled by tree walker.
+
+		case T_CreateTableAsStmt: {
+			CreateTableAsStmt* stmt = castNode(CreateTableAsStmt, node);
+			if (stmt->into && stmt->into->rel)
+				add_from_clause_item(stmt->into->rel, CONTEXT_DDL, state);
+
+			if (stmt->query)
+				pg_query_summary_walk_impl(stmt->query, state);
```
lfittl:
```suggestion
			pg_query_summary_walk_impl(stmt->query, state);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+
+			ListCell* lc = NULL;
+			foreach(lc, stmt->indexParams) {
+				IndexElem* index_elem = lfirst(lc);
+				if (index_elem->expr)
+					pg_query_summary_walk_impl(index_elem->expr, state);
+			}
+
+			if (stmt->whereClause)
+				pg_query_summary_walk_impl(stmt->whereClause, state);
```
lfittl:
```suggestion
			pg_query_summary_walk_impl(stmt->whereClause, state);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		case T_IndexStmt: {
+			IndexStmt* stmt = castNode(IndexStmt, node);
+
+			add_from_clause_item(stmt->relation, CONTEXT_DDL, state);
+
+			ListCell* lc = NULL;
+			foreach(lc, stmt->indexParams) {
+				IndexElem* index_elem = lfirst(lc);
+				if (index_elem->expr)
+					pg_query_summary_walk_impl(index_elem->expr, state);
```
lfittl:
```suggestion
				pg_query_summary_walk_impl(index_elem->expr, state);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		}
+
+		case T_CreateTrigStmt:
+			add_from_clause_item(castNode(CreateTrigStmt, node)->relation, CONTEXT_DDL, state);
+			return false; // Not handled by tree walker.
+
+		case T_RuleStmt: {
+			add_from_clause_item(castNode(RuleStmt, node)->relation, CONTEXT_DDL, state);
+			return false; // Not handled by tree walker.
+		}
```
lfittl:
```suggestion
		case T_RuleStmt:
			add_from_clause_item(castNode(RuleStmt, node)->relation, CONTEXT_DDL, state);
			return false; // Not handled by tree walker.
```

(since it doesn't define a variable)

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+			}
+
+			return false; // Not handled by tree walker.
+		}
+
+		case T_RefreshMatViewStmt:
+			add_from_clause_item(castNode(RefreshMatViewStmt, node)->relation, CONTEXT_DDL, state);
+			return false; // Not handled by tree walker.
+
+		case T_FuncCall:
```
lfittl: I would move this down right before `T_A_Expr` for consistency (since this is not part of the original tree walk in the Ruby implementation)

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+					foreach(lc, stmt->objects) {
+						Node* obj = lfirst(lc);
+						if (obj->type == T_List) {
+							List* list = castNode(List, obj);
+
+							size_t len = list_length(list);
+							if (len == 0)
+								continue;
+
+							if (len == 1)
```
lfittl: We may also have do deal with `len = 3` here, in case of a table qualified by both schema and database name (`DROP TABLE db.schema.x`)

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+	ListCell* lc = NULL;
+	foreach(lc, state->cte_names) {
+		char *ctename = lfirst(lc);
+		if (strcmp(ctename, relname) == 0)
+			return true;
+	}
+	return false;
+}
+
+static void add_table(WalkState* state, ContextType context, char* schemaname, char* relname) {
```
lfittl: We should probably locate this close to `add_function`, since they behave similarly.

lfittl: Also, if this takes a `RangeVar` directly, we could simplify the calling code.

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		case T_DropStmt: {
+			DropStmt* stmt = castNode(DropStmt, node);
+
+			switch (stmt->removeType) {
+				case OBJECT_TABLE: {
+					ListCell* lc = NULL;
+					foreach(lc, stmt->objects) {
+						Node* obj = lfirst(lc);
+						if (obj->type == T_List) {
+							List* list = castNode(List, obj);
```
lfittl: If we had [makeRangeVarFromNameList](https://github.com/postgres/postgres/blob/88f0fdabead51539aa5bdefcf188f07d4651ee10/src/backend/catalog/namespace.c#L3554), we could refactor this:

```suggestion
							List* list = castNode(List, obj);
							if (list_length(list) == 0)
								continue;
							RangeVar *range_var = makeRangeVarFromNameList(castNode(List, obj));
							add_table(state, CONTEXT_DDL, range_var);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+		if (list_cell_number(funcname, lc) != 0)
+			appendStringInfoChar(&buf, '.');
+
+		String* str = lfirst_node(String, lc);
+		appendStringInfoString(&buf, quote_identifier(str->sval));
+	}
+
+	SummaryFunction* func = palloc(sizeof(SummaryFunction));
+
+	func->name = buf.data;
```
lfittl: We can simplify this with [NameListToString](https://github.com/postgres/postgres/blob/master/src/backend/catalog/namespace.c#L3594):
```suggestion
	SummaryFunction* func = palloc(sizeof(SummaryFunction));

	func->name = NameListToString(funcname);
```

src/pg_query_summary.c
lfittl: (this would resolve itself if we switch to makeRangeVarFromNameList)

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+								continue;
+
+							if (len == 1) {
+								String* relname = linitial(list);
+								add_table(state, CONTEXT_DDL, NULL, relname->sval);
+							} else if (len == 2) {
+								String* schemaname = linitial(list);
+								String* relname = lsecond(list);
+								add_table(state, CONTEXT_DDL, schemaname->sval, relname->sval);
+							}
```
lfittl:
```suggestion
							// Ignore the last string (the rule/trigger name)
							List* list = list_copy(castNode(List, obj));
							list = list_truncate(list, list_length(list) - 1);

							if (list_length(list) == 0)
								continue;
							RangeVar *range_var = makeRangeVarFromNameList(castNode(List, obj));
							add_table(state, CONTEXT_DDL, range_var);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,761 @@
...
+					break;
+				}
+
+				default:
+					break;
+			}
+			return false; // Not handled by tree walker.
+		}
+
+		case T_GrantStmt: {
```
lfittl: Note for next review session: continue here.

duckinator:
Currently, `makeRangeVarFromNameList()` is unceremoniously thrown in the top of `pg_query_summary.c`. We should import it _properly_ like other Postgres functions are, then remove that.

Aside from that and reformatting curly braces, all of your review feedback should be handled.

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			// Ruby just says "FIXME" for both, with no context provided.
+			// https://github.com/pganalyze/pg_query/blob/main/lib/pg_query/parse.rb
+			//
+			// Rust doesn't even acknowledge anything but OBJECT_TABLE.
+			if (stmt->objtype == OBJECT_TABLE) {
+				ListCell* lc = NULL;
+				foreach(lc, stmt->objects) {
+					add_from_clause_item(lfirst(lc), CONTEXT_DDL, state);
+				}
+			}
```
lfittl:
```suggestion
			switch (stmt->objtype) {
			case OBJECT_TABLE:
			    {
				ListCell* lc = NULL;
				foreach(lc, stmt->objects) {
					add_from_clause_item(lfirst(lc), CONTEXT_DDL, state);
				}
				break;
			}
			case OBJECT_FUNCTION, OBJECT_PROCEDURE, OBJECT_ROUTINE:
			{
				ListCell* lc = NULL;
				foreach(lc, stmt->objects) {
					if (obj_node != NULL && obj_node->type == T_ObjectWithArgs) {
						ObjectWithArgs* obj = castNode(ObjectWithArgs, lfirst(lc));

						if (obj->objname)
							add_function(obj->objname, CONTEXT_DDL, state);
					}
				}
				break;
			}
			default:
			  /* Do nothing, other grant types do not contain table or function references. */
			}
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			return false; // Not handled by tree walker.
+		}
+
+		case T_LockStmt: {
+			LockStmt* stmt = castNode(LockStmt, node);
+
+			ListCell* lc = NULL;
+			foreach(lc, stmt->relations) {
+				add_from_clause_item(lfirst(lc), CONTEXT_DDL, state);
+			}
```
lfittl:
```suggestion
			add_from_clause_items(stmt->relations, CONTEXT_DDL, state);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			add_function(castNode(CreateFunctionStmt, node)->funcname, CONTEXT_DDL, state);
+			return false; // Not handled by tree walker.
+
+		case T_RenameStmt: {
+			RenameStmt* stmt = castNode(RenameStmt, node);
+			if (stmt->renameType == OBJECT_FUNCTION) {
+				ObjectWithArgs* obj = castNode(ObjectWithArgs, stmt->object);
+				add_function(obj->objname, CONTEXT_DDL, state);
+				add_function(list_make1(makeString(stmt->newname)), CONTEXT_DDL, state);
+			}
```
lfittl:
```suggestion
			switch (stmt->renameType)
			{
			case OBJECT_TABLE:
			{
				add_from_clause_item(stmt->relation, CONTEXT_DDL, state);
				add_from_clause_item(makeRangeVar(stmt->relation->schemaname, stmt->newname, -1), CONTEXT_DDL, state);
			}
			case OBJECT_FUNCTION, OBJECT_PROCEDURE, OBJECT_ROUTINE:
			{
				ObjectWithArgs* obj = castNode(ObjectWithArgs, stmt->object);
				add_function(obj->objname, CONTEXT_DDL, state);
				add_function(list_make1(makeString(stmt->newname)), CONTEXT_DDL, state);
				break;
			}
			}
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			foreach(lc, stmt->relations) {
+				add_from_clause_item(lfirst(lc), CONTEXT_DDL, state);
+			}
+
+			return false; // Not handled by tree walker.
+		}
+
+		case T_ExplainStmt:
+			pg_query_summary_walk_impl(castNode(ExplainStmt, node)->query, state);
+			return false; // Not handled by tree walker.
```
lfittl:
```suggestion
		case T_ExplainStmt:
			return pg_query_summary_walk_impl(castNode(ExplainStmt, node)->query, state);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			add_function(castNode(FuncCall, node)->funcname, CONTEXT_CALL, state);
+			break;
+
+		// The following are added to avoid needing subselect_items,
+		// like is used in Rust and Ruby.
+
+		case T_A_Expr: {
+			A_Expr* expr = castNode(A_Expr, node);
+			pg_query_summary_walk_impl(expr->lexpr, state);
+			pg_query_summary_walk_impl(expr->rexpr, state);
```
lfittl:
```suggestion
			if (pg_query_summary_walk_impl(expr->lexpr, state))
				return true;
			if (pg_query_summary_walk_impl(expr->rexpr, state))
				return true;
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+			return pg_query_summary_walk_impl(castNode(RawStmt, node)->stmt, state);
+
+		case T_FuncCall:
+			add_function(castNode(FuncCall, node)->funcname, CONTEXT_CALL, state);
+			break;
+
+		// The following are added to avoid needing subselect_items,
+		// like is used in Rust and Ruby.
+
+		case T_A_Expr: {
```
lfittl: This may not be needed since the raw expression tree walker does handle it (https://github.com/postgres/postgres/blob/d699687b329e031cd90e967b39c3fd8a53ef8208/src/backend/nodes/nodeFuncs.c#L4388)

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+
+			ListCell* lc = NULL;
+			foreach(lc, ref->fields) {
+				Node* field_node = lfirst(lc);
+
+				// If it's not a String, skip it.
+				if (field_node && field_node->type != T_String)
+					continue;
+
+				String* field = castNode(String, field_node);
```
lfittl:
```suggestion
				// If it's not a String, skip it.
				if (!IsA(lfirst(lc), String))
					continue;

				String* field = lfirst_node(String, lc);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+
+			return false; // Not handled by tree walker.
+		}
+
+		case T_ColumnRef: {
+			if (!state->save_filter_columns)
+				break;
+
+			ColumnRef* ref = castNode(ColumnRef, node);
+
```
lfittl, line 500:
```
Node *column_node = llast(ref->fields);
/* Check that we are not in the A_Star case here (columns can be stars, nothing else can be) */
if (!IsA(column, String))
  break;

String* schema_name = NULL;
String* table_name = NULL;
String* column = castNode(String, column_node);
if (list_length(ref->fields) == 2)
{
  table_name = linitial(ref->fields);
} else if (list_length(ref->fields) == 3)
{
  schema_name = linitial(ref->fields);
  table_name = list_nth_cell(ref->fields, 1);
}
```
(there may be subtle bugs, but this should be easier to work with)

src/pg_query_summary.c
lfittl: (or may not be needed if we avoid the loop)

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+		case T_TransactionStmt:
+		case T_VariableShowStmt:
+		case T_VariableSetStmt:
+			return false;
+
+		default:
+			break;
+	}
+
+	return raw_expression_tree_walker(node, pg_query_summary_walk_impl, (void*) state);
```
lfittl: We stopped review here.

src/pg_query_summary.c
duckinator: I inferred that `obj_node` was _probably_ intended to be defined something like I mention below. It'd be a good idea to look over the updated code to double-check that it's as you intended.

```
foreach(lc, stmt->objects) {
    Node* obj_node = lfirst(lc)
    if (obj_node != NULL && obj_node->type == T_ObjectWithArgs) {
        ObjectWithArgs* obj = castNode(ObjectWithArgs, obj_node);
        // ...
    }
}
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+				}
+				case OBJECT_FUNCTION:
+				case OBJECT_PROCEDURE:
+				case OBJECT_ROUTINE:
+				{
+					ListCell* lc = NULL;
+					foreach(lc, stmt->objects) {
+						Node* obj_node = lfirst(lc);
+
+						if (obj_node != NULL && obj_node->type == T_ObjectWithArgs)
```
lfittl:
```suggestion
						if (IsA(obj_node, ObjectWithArgs))
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+			return false;
+
+		default:
+			break;
+	}
+
+	return raw_expression_tree_walker(node, pg_query_summary_walk_impl, (void*) state);
+}
+
+static char* range_var_parts_to_name(char* schemaname, char* relname) {
```
lfittl: There doesn't appear to be an existing function in Postgres that takes care of this, but we could simplify this a bit and have it take a `RangeVar` directly (but no strong opinions on that).

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+// Walk the parse tree, storing the results in `summary`.
+//
+// Returns NULL on success.
+PgQueryError* pg_query_summary_walk(Summary* summary, Node* tree) {
+	WalkState state = {0};
+
+	// NOTE regarding cte_names and from_clause_items:
+	// 1. We process cte_names as we iterate through the tree.
+	// 2. We only add items to tables if they are *not* in cte_names.
+	// 3. CTEs can be referenced *after* table names.
```
lfittl:
```suggestion
	// 3. CTEs can be defined *after* they're referenced as table names.
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+static void add_from_clause_item(Node* node, ContextType context, WalkState* state) {
+	if (!node)
+		return;
+
+	NodeWithContext* nwc = palloc(sizeof(NodeWithContext));
+	nwc->node = node;
+	nwc->context = context;
+
+	state->from_clause_items = lappend(state->from_clause_items, nwc);
+}
```
lfittl:
```suggestion
static void add_range_var_with_context(Node* node, ContextType context, WalkState* state) {
	if (!node)
		return;
	
	if (IsA(node, JoinExpr))
	{
	   add_range_var_with_context(node->larg, context, state);
	   add_range_var_with_context(node->rarg, context, state);
	   // Note that quals is handled by the regular tree walk in the off-chance it contains a SelectStmt
	   return;
	}

	Assert(IsA(node, RangeVar));

	RangeVarWithContext* rwc = palloc(sizeof(RangeVarWithContext));
	nwc->range_var = (RangeVar *) node;
	nwc->context = context;

	state->range_vars_with_context = lappend(state->range_vars_with_context, rwc);
}
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+					foreach(lc, stmt->objects) {
+						Node* obj = lfirst(lc);
+						if (obj->type == T_List) {
+							List* list = castNode(List, obj);
+
+							if (list_length(list) == 0)
+								continue;
+
+							RangeVar* range_var = makeRangeVarFromNameList(list);
+							add_table(state, CONTEXT_DDL, range_var);
```
lfittl: For consistency, we can use (the now renamed) `add_range_var_with_context` here as well:
```suggestion
							add_range_var_with_context(state, CONTEXT_DDL, makeRangeVarFromNameList(list));
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+			handle_from_clause_item((Node*) je->larg, context, state);
+			handle_from_clause_item((Node*) je->rarg, context, state);
+
+			handle_from_clause_item(je->quals, context, state);
+
+			break;
+		}
+		case T_RangeVar: {
+			RangeVar* rv = castNode(RangeVar, node);
+			add_table(state, context, castNode(RangeVar, node));
```
lfittl: With this method simplified to just handle RangeVars we could inline `add_table` here for clarity and to discourage anyone from calling it directly.

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+
+	pg_query_summary_walk_impl(tree, &state);
+
+	ListCell* lc = NULL;
+	foreach(lc, state.from_clause_items) {
+		NodeWithContext* lc_node_with_context = lfirst(lc);
+
+		Node* lc_node = lc_node_with_context->node;
+
+		handle_from_clause_item(lc_node, lc_node_with_context->context, &state);
```
lfittl: Per the other notes, if we special case this to just be range vars, it may also be nice to simplify the call here to pass the `RangeVarWithContext` directly and have the (renamed) `handle_range_var_with_context` take that.

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+		summary->cte_names = state.cte_names;
+
+	if (state.functions != NULL)
+		summary->functions = state.functions;
+
+	if (state.filter_columns != NULL)
+		summary->filter_columns = state.filter_columns;
+
+	// No error to return.
+	return NULL;
```
lfittl: I think we can make this function `void` since it never returns an error.

src/pg_query_summary.c
```diff
@@ -0,0 +1,775 @@
...
+	return make_pg_query_error("summary_truncate() is not implemented yet");
+
+	// TODO: implement summary_truncate()
+
+	// No error to return.
+	return NULL;
+}
+
+// Given the summarized query, convert it to protobuf, and store it in `result`.
+static void summary_to_protobuf(PgQuerySummaryParseResult* result, Summary* summary)
```
lfittl: Need to do a quick look over this in terms of memory allocation logic next time.

duckinator:
The change from `add_from_clause_item()` to `add_range_var_with_context()` revealed something unexpected: the smaller switch statement in `handle_from_clause_item()` was getting &mdash; and completely dropping &mdash; `RangeFunction` and `RangeSubselect` objects.

I don't know if those need to be handled or not. If they do, I'm not sure how to do so, since they seem to contain different information than a `RangeVar`.

lfittl:
> The change from `add_from_clause_item()` to `add_range_var_with_context()` revealed something unexpected: the smaller switch statement in `handle_from_clause_item()` was getting — and completely dropping — `RangeFunction` and `RangeSubselect` objects.
> 
> I don't know if those need to be handled or not. If they do, I'm not sure how to do so, since they seem to contain different information than a `RangeVar`.

Hmm, that's interesting. Do you have an example query that produces them?

duckinator:
`SELECT * FROM my_custom_func()` (`it_finds_functions_in_FROM_clause`) causes it to get a `RangeFunction`.

`select u.* from (select * from users) u` (`it_finds_nested_tables_in_FROM_clause`) causes it to get a `RangeSubselect`.

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+		}
+
+		default:
+			elog(ERROR, "unexpected node type: %d", nodeTag(node));
+	}
+}
+
+// Given a List*-of-Nodes (clause) and WalkState (state), wrap them all
+// in RangeVarWithContext structs and add them to the state->range_vars list.
+static void add_range_vars_with_context(List *clause, ContextType context, WalkState *state) {
```
lfittl: As mentioned, we could simplify this by moving this into a `T_List` branch in the switch statement of `add_range_var_with_context`, so we have a single function that's being called.

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+	}
+	return false;
+}
+
+static void handle_range_var(RangeVar* node, ContextType context, WalkState* state) {
+	if (node == NULL)
+		return;
+
+	RangeVar* rv = node;
+	add_table(state, context, castNode(RangeVar, node));
```
lfittl: We can inline this here since this is the only caller of `add_table`.

Makefile
```diff
@@ -269,6 +269,7 @@ else
 	test/normalize
 	test/normalize_utility
 	test/parse
+	test/summary
```
lfittl: Need to also add this to the valgrind list, as well as `TESTS` so the binary is built correctly.

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+}
+
+// Given the summarized query, convert it to protobuf, and store it in `result`.
+static void summary_to_protobuf(PgQuerySummaryParseResult* result, Summary* summary)
+{
+	PgQueryProtobuf protobuf;
+	PgQuery__SummaryResult sr = PG_QUERY__SUMMARY_RESULT__INIT;
+
+	sr.n_tables = list_length(summary->tables);
+	sr.tables = malloc(sizeof(PgQuery__SummaryResult__Table*) * sr.n_tables);
```
lfittl: We should be able to use `palloc`/`pfree` in this function, with the exception of the data pointer that's returned, which has to use `malloc` (per the comment there).

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+	foreach(lc, summary->tables) {
+		SummaryTable* table = lfirst(lc);
+		sr.tables[i] = malloc(sizeof(PgQuery__SummaryResult__Table));
+		pg_query__summary_result__table__init(sr.tables[i]);
+		sr.tables[i]->name = table->name;
+		sr.tables[i]->schema_name = table->schema_name;
+		sr.tables[i]->table_name = table->table_name;
+		sr.tables[i]->context = (PgQuery__SummaryResult__Context)table->context;
+		i++;
+	}
```
lfittl:
```suggestion
	ListCell *lc = NULL;
	foreach(lc, summary->tables) {
	    size_t i = foreach_current_index(lc);
		SummaryTable* table = lfirst(lc);
		sr.tables[i] = malloc(sizeof(PgQuery__SummaryResult__Table));
		pg_query__summary_result__table__init(sr.tables[i]);
		sr.tables[i]->name = table->name;
		sr.tables[i]->schema_name = table->schema_name;
		sr.tables[i]->table_name = table->table_name;
		sr.tables[i]->context = (PgQuery__SummaryResult__Context)table->context;
	}
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+
+	sr.n_aliases = list_length(summary->aliases);
+	sr.aliases = malloc(sizeof(PgQuery__SummaryResult__AliasesEntry*) * sr.n_aliases);
+	lc = NULL;
```
lfittl: In my understanding `lc` should be reset in the `foreach`macro already, so we can remove this.

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+	}
+
+	sr.n_statement_types = list_length(summary->statement_types);
+	sr.statement_types = malloc(sizeof(char*) * sr.n_statement_types);
+	lc = NULL;
+	i = 0;
+	foreach(lc, summary->statement_types) {
+		char* st = lfirst(lc);
+
+		sr.statement_types[i] = strdup(st);
```
lfittl:
```suggestion
		sr.statement_types[i] = pstrdup(st);
```

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+
+	if (result.error == NULL)
+		summary_to_protobuf(&result, &summary);
+
+	pg_query_exit_memory_context(direct_result.ctx);
+
+	return result;
+}
+
+// FIXME: Reduce duplication with non-_direct variant?
```
lfittl: I think that's not worth it, better if we keep the slight duplication here.

src/pg_query_summary.h
```diff
@@ -0,0 +1,90 @@
...
+typedef struct {
+	char *schema_name;
+	char *table_name;
+	char *column;
+} FilterColumn;
+
+typedef struct {
+	List *tables; // List of SummaryTable*
+	List *aliases; // List of SummaryAlias*
+	List *cte_names;
```
lfittl:
```suggestion
	List *cte_names; // List of char*
```

src/pg_query_summary.h
```diff
@@ -0,0 +1,90 @@
...
+	List *statement_types; // List of char*
+
+	// UNIMPLEMENTED(truncated_query): char *truncated_query;
+} Summary;
+
+typedef struct {
+	List* tables; // List of Table*
+	List *aliases; // List of SummaryAlias*
+	List* functions; // List of Function*
+	List* subselect_items;
```
lfittl: I think we can drop this, yay!

src/pg_query_summary.h
```diff
@@ -0,0 +1,90 @@
...
+	List* range_vars;
```
lfittl:
```suggestion
	List* range_vars; // List of RangeVarWithContext*
```

src/pg_query_summary.h
```diff
@@ -0,0 +1,90 @@
...
+} PgQuerySummaryParseResultDirect;
+
+PgQueryError* pg_query_summary_statement_walk(Summary *summary, Node *node);
+
+// This is like pg_query_summary(), but returns a (non-protobuf)
+// PgQuerySummaryParseResultDirect instead of the usual
+// PgQuerySummaryParseResult which converts the summary to protobuf.
+//
+// This is used for tests, to avoid having to process protobuf in C.
+PgQuerySummaryParseResultDirect pg_query_summary_direct(const char* input, int parser_options, int truncate_limit);
```
lfittl: Maybe `_internal` and `PgQuerySummaryParseResultInternal` would be better, to make it clear to the caller we don't expect them to use it.

src/pg_query_summary_statement_type.c
```diff
@@ -0,0 +1,798 @@
...
+#include "utils/builtins.h"
+
+#include <unistd.h>
+#include <fcntl.h>
+#include <string.h>
+
+#include "pg_query_summary.h"
+
+// Declare the huge function early so we can put it at the end of the file.
+bool pg_query_summary_stmt_walk_impl(Node* node, Summary* summary);
```
lfittl: I think this function could use a `static` (both in this early declaration and the later definition)

src/pg_query_summary_statement_type.c
```diff
@@ -0,0 +1,798 @@
...
+	}
+
+	summary->statement_types = lappend(summary->statement_types, new_stmt_type);
+}
+
+
+PgQueryError* pg_query_summary_statement_walk(Summary* summary, Node* tree) {
+	pg_query_summary_stmt_walk_impl(tree, summary);
+
+	return NULL;
```
lfittl: We can drop this and make the function a `void` return type.

src/pg_query_summary_statement_type.c
```diff
@@ -0,0 +1,798 @@
...
+			add_stmt_type_if_needed(summary, "AlterTableSpaceOptionsStmt");
+			break;
+		case T_AlterTableStmt: {
+			add_stmt_type_if_needed(summary, "AlterTableStmt");
+			// Return early because raw_expression_tree_walker doesn't handle this.
+			return false;
+		}
+		case T_AlterTSConfigurationStmt:
+			// Capitalization is deliberately different to match Rust implementation.
+			add_stmt_type_if_needed(summary, "AlterTsconfigurationStmt");
```
lfittl: I think we should intentionally break from the Rust implementation here - this was not an intentional choice (to my knowledge), and it'd be good to exactly match the C struct names.

src/pg_query_summary_statement_type.c
```diff
@@ -0,0 +1,798 @@
...
+
+	return NULL;
+}
+
+bool pg_query_summary_stmt_walk_impl(Node* node, Summary* summary) {
+	if (node == NULL) return false;
+
+	switch(nodeTag(node)) {
+		case T_AlterCollationStmt:
+			add_stmt_type_if_needed(summary, "AlterCollationStmt");
```
lfittl, line 57: I think all of the utility statements here (i.e. things that are not SELECT/UPDATE/DELETE/INSERT/MERGE) are not handled by the tree walker, so we should return early. You can check https://github.com/postgres/postgres/blob/master/src/backend/nodes/nodeFuncs.c#L4000 to confirm.

src/pg_query_summary_statement_type.c
```diff
@@ -0,0 +1,798 @@
...
+		case T_XidList:
+		case T_AllocSetContext:
+		case T_GenerationContext:
+		case T_SlabContext:
+		case T_BumpContext:
+		case T_TIDBitmap:
+		case T_WindowObjectData:
+			break;
+	}
+
```
lfittl, line 792: If we invented a function that checks whether a given node is handled by the treewalker, we can call that here, and `return false` if its not supported (that way the code above is simplified, and we can share code between `pg_query_summary.c` and `pg_query_summary_statement.c`)

src/pg_query_summary.c
```diff
@@ -0,0 +1,780 @@
...
+bool pg_query_summary_walk_impl(Node* node, WalkState* state) {
+	if (node == NULL)
+		return false;
+
+	// NOTE: If `raw_expression_tree_walker()` doesn't handle a particular
+	// node type, you will need to `return false` early to avoid passing it
+	// to the tree walker.
+	// If you do this, be sure to update `pg_query_summary_stmt_walk_impl()`.
+
+	switch (nodeTag(node)) {
```
lfittl: There are certain utility statements not yet handled here, for example `AlterExtensionStmt`. Because the treewalker doesn't support those, it would cause an error - as discussed in `pg_query_summary_statement.c` maybe easiest to deal with by introducing a special function for checking whether a node type is not supported by the raw expression tree walker (and then returning false before we call the treewalker at the end of this function).

This will also let us consistently use `break;` in this switch statement without having to `return false`.

src/pg_query_summary.c
duckinator: I couldn't get this working because of type issues. However, I did rename the two functions, which at least makes it a lot less likely to confuse the two:

- `add_range_var_with_context()` => `add_range_var()`
- `add_range_vars_with_context()` => `add_range_var_list()`

src/pg_query_raw_tree_walker_supports.c
```diff
@@ -0,0 +1,116 @@
...
+#include "nodes/nodeFuncs.h"
+#include "nodes/pathnodes.h"
+
+#include "utils/builtins.h"
+
+// Returns true if raw_expression_tree_walker() handles it,
+// otherwise, returns false.
+//
+// Up-to-date with postgres' implementation as of Aug 19 2025.
+// https://github.com/postgres/postgres/blob/master/src/backend/nodes/nodeFuncs.c#L4000
```
lfittl: Its probably the same, but we should make this match the state of the `REL_17_STABLE` (which is the Postgres 17 released version, which the current libpg_query is based on), instead of `master` (which is the Postgres 19 development branch)

src/pg_query_raw_tree_walker_supports.c
```diff
@@ -0,0 +1,116 @@
...
+		case T_Integer:
+		case T_Float:
+		case T_Boolean:
+		case T_String:
+		case T_BitString:
+		case T_ParamRef:
+		case T_A_Const:
+		case T_A_Star:
+		case T_MergeSupportFunc:
+		//case T_ReturningOption:
```
lfittl: We could leave these in and state `TODO: Add this back for Postgres 18`

src/pg_query_summary.h
```diff
@@ -0,0 +1,91 @@
...
+// This is like PgQuerySummaryParseResult, but the summary is not wrapped
+// up in protobuf.
+typedef struct {
+	Summary summary;
+	char* stderr_buffer;
+	PgQueryError* error;
+	MemoryContext ctx;
+} PgQuerySummaryParseResultInternal;
+
+bool pg_query_raw_tree_walker_supports(Node* node);
```
lfittl: We should add `extern` keywords in front of the functions defined in the header file for consistency.

duckinator:
I've been looking into valgrind stuff, it looks like this is what's going on:

1. `test/summary.c` calls `pg_query_enter_memory_context()` (for `lappend()` usage), which:
  a. calls `pg_query_init()` which calls `MemoryContextInit()` which sets `TopMemoryContext`.
  b. calls `Assert(CurrentMemoryContext == TopMemoryContext)` (this is true)
  c. calls `ctx = AllocSetContextCreate(...)`
  d. calls `MemoryContextSwitchTo(ctx)` which changes `CurrentMemoryContext`
2. `test/summary.c` calls `pg_query_summary_internal(...)` which calls `pg_query_enter_memory_context()` again, which means
  a. calls `pg_query_init()`, which returns immediately because it was already initialized
  b. calls `Assert(CurrentMemoryContext == TopMemoryContext)` &mdash; this is **false** now
  c. calls `ctx = AllocSetContextCreate(...)`
  d. calls `MemoryContextSwitchTo(ctx)` which changes `CurrentMemoryContext`

So this is very clearly breaking an existing assumption.

~~It's unclear to me whether the Assert() (and thus the assumption) is wrong, or if it really can't handle 3-deep memory contexts.~~

**EDIT:** Given that `pg_query_exit_memory_context(ctx)` is hard-coded to switch to to `TopLevelContext` and refers to that as the "previous" memory context, I think the assumption is correct.

So I need to tweak my tests to avoid indirectly `palloc()`ing things.

src/pg_query_summary.c
```diff
@@ -0,0 +1,765 @@
...
+		size_t i = foreach_current_index(lc);
+		SummaryFunction* fn = lfirst(lc);
+
+		sr.functions[i] = palloc(sizeof(PgQuery__SummaryResult__Function));
+		pg_query__summary_result__function__init(sr.functions[i]);
+		sr.functions[i]->name = fn->name;
+		sr.functions[i]->function_name = fn->function_name;
+		sr.functions[i]->schema_name = fn->schema_name;
+		// FIXME: Implicit cast from enum ContextType to enum _PgQuery__SummaryResult__Context.
+		// This can't actually be guaranteed as correct -- probably needs actual logic, sadly.
```
lfittl: As discussed, this would be good to take care of now, so we don't forget about it. Since this is not performance critical having explicit mapping logic with a switch statement seems fine.

test/summary.c
```diff
@@ -0,0 +1,214 @@
...
+	size_t failed;
+	size_t skipped;
+	MemoryContext last_ctx;
+} TestSuiteState;
+
+static TestSuiteState test_state = {0};
+
+typedef void (TestFn)(void);
+
+#define TEST_INIT() printf("%s (line %i)\n", __func__, __LINE__)
```
lfittl: As discussed, if we can make it work to have this in a helper file (`.c` or `.h`) so its reusable, that'd be great.

Makefile
```diff
@@ -241,7 +241,7 @@ examples/normalize_error: examples/normalize_error.c $(ARLIB)
 examples/simple_plpgsql: examples/simple_plpgsql.c $(ARLIB)
```
lfittl, line 237: Earlier in this file, we need to modify a piece of code, so that tests correctly run on mingw on Windows:

```
ifeq ($(OS),Windows_NT)
override CFLAGS += -I./src/postgres/include/port/win32
override TEST_CFLAGS += -I./src/postgres/include/port/win32
endif
```

Also, we need to add these tests to `Makefile.msvc`.

test/summary_tests.c
```diff
@@ -0,0 +1,1038 @@
...
+	// UNIMPLEMENTED(truncated_query): assert_eq!(result.truncated_query.is_none(), true);
+	TEST_ASSERT_LIST_EQUAL(result.statement_types, list_make1("SelectStmt"));
+}
+
+void it_parses_simple_query_with_alias(TestState* test_state) {
+	TEST_INIT();
+	Summary result = summary("SELECT * FROM test AS x WHERE a = 1", 0, -1);
+
+	TEST_ASSERT_LIST_LENGTH(result.tables, 1);
+	TEST_SUMMARY_ASSERT_TABLES(result.tables, ((char*[]){"test", NULL}));
```
lfittl, line 20: As discussed, we should also check for context here - maybe we can have TEST_SUMMARY_ASSERT_TABLES take the context it should check for? (SELECT vs DDL, etc)

Same goes for function names.

duckinator:
As a note for myself, in case I need to back out the rebase for any reason: Current HEAD on my `summary` branch is 3885a3f.

lfittl (approved):
:+1: I think this is ready to merge :)

duckinator:
@lfittl this looks good to me! There's some minor cleanup that can be done, such as not having a `test_cleanup()` at all and doing `test_run_with_mctx(tests, NULL)`, but given how obscenely complex this PR has gotten I would like to _actively encourage_ that kind of stuff being a follow-up.

PR #278, opened 2025-04-02
Release 17-6.1.0

lfittl:


Makefile
```diff
@@ -13,7 +13,7 @@ PG_VERSION_NUM = 170004
 PG_BRANCH = REL_17_STABLE
 PROTOC_VERSION = 25.1
 
-VERSION = 5.1.0
```
lfittl, line 16: Note we missed updating this when we released 6.0.0, but because this is mostly not used (only for edge cases where someone wants to build a shared library) we did not notice until now.

keiko713 (approved):

seanlinsley:
There's a new bug report that we may want to address first https://github.com/pganalyze/libpg_query/issues/280

lfittl:
> There's a new bug report that we may want to address first #280

Thanks, I fixed that in this commit added to the PR: https://github.com/pganalyze/libpg_query/pull/278/commits/2abfce11efe7c2fdaf253de3d6cee4d37d5f3a76

PR #288, opened 2025-06-06
Add support for normalizing CALL statements

msakrejda:
See also https://github.com/pganalyze/pg_query/pull/336


msakrejda:
It looks like some jobs were skipped because the CI OS has been phased out. I updated that in #289.

lfittl (approved):

PR #289, opened 2025-06-06
Update GitHub Actions Linux OS to 24.04

msakrejda:
20.04 has been removed from GitHub Actions.


msakrejda:
Hmm. Not sure what's up with the failures.

lfittl:
> Hmm. Not sure what's up with the failures.

Odd - I wonder if its a disk space problem? Might be worth trying with larger runners and see if that changes it.

msakrejda:
Are there runners with more disk space? [This](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) seems to suggest that they all have 14GB.

lfittl:
> Are there runners with more disk space? [This](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) seems to suggest that they all have 14GB.

Yep, the "large runners" do, see https://docs.github.com/en/actions/concepts/runners/about-larger-runners

e.g. you could try `ubuntu-22.04-4-cores`

msakrejda:
Yep, just found that :smile: . Trying it now.

msakrejda:
Hmm, same failures: https://github.com/pganalyze/libpg_query/actions/runs/15931693704/job/44942139688?pr=289#step:8:1913

lfittl:
> Hmm, same failures: https://github.com/pganalyze/libpg_query/actions/runs/15931693704/job/44942139688?pr=289#step:8:1913

Hmm, strange. That seems like a valgrind-specific problem then. I can dig into this a bit later.

msakrejda:
Could it be https://github.com/llvm/llvm-project/issues/56550? Same error message, someone on the thread claims it's fixed in 3.20, and it looks like the latest valgrind on 22.04 is 3.18.1. So I think we either need to disable the clang/valgrind combo in CI, or update the OS to 24.04. Or we could add 24.04 to the matrix (and maybe only enable valgrind on one OS per other matrix combination as a compromise).

msakrejda:
I tried 24.04 but it's hitting [a different error](https://github.com/pganalyze/libpg_query/actions/runs/15937946538/job/44961425841?pr=289#step:8:84).

lfittl:
> I tried 24.04 but it's hitting [a different error](https://github.com/pganalyze/libpg_query/actions/runs/15937946538/job/44961425841?pr=289#step:8:84).

Interesting. Looks like that did succeed on everything except clang with protobuf-ccp and valgrind. I'll restart, just to see if by chance this fixes itself when re-running.

.github/workflows/ci.yml
```diff
@@ -21,7 +26,7 @@ jobs:
       uses: actions/cache@v4
       with:
         path: protobuf-25.1
-        key: ${{ runner.os }}-protobuf-25.1
+        key: ${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.valgrind }}-protobuf-25.1
```
msakrejda: Should `valgrind` be in the cache key, or no? I'm not actually sure what exactly we're caching here.

lfittl: I think you can drop it.

lfittl (approved):

.github/workflows/ci.yml
msakrejda:
```suggestion
        key: ${{ runner.os }}-${{ matrix.compiler }}-protobuf-25.1
```

PR #293, opened 2025-07-02
Deparser: Add optional formatting/pretty printing

lfittl:
## TODO

* [x] Review UNION handling (see complex_mattm.sql test case)
* [x] Add additional test cases
* [x] Review outputs for typical pg_dump
* [x] Review comment handling (especially in regards to how the API is structured)
* [x] Consider JOINs like other FROM elements (indent them like table names in FROM)
* [x] Keep certain fields on one line (e.g. SELECT target list) up to a certain line length limit
* [x] Introduce libpg_query convenience function for extracting comments from a source query (so we don't replicate this across the different bindings)
* [x] Add option to emit commas at start of a line? (to help with reducing diffs in commits that change a command, like the INSERT at the end of Rails' `db/structure.sql`)
* [x] Review TODO items mentioned in tests
* [x] Make sure "make extract_source" pulls in the necessary functions (some were manually added to fix issues)
* [x] Figure out how to pass comment emission state across statements (see comment from code review)
* [x] Add code comment explaining parts/part groups/nesting levels (and specifically note how increase/decrease works with nesting levels and that they are supposed to be paired within a function)

Defer until later

* Improve handling of long lists generated by `deparseExprList` (e.g. see `21-in.psql` and `03-many-columns.psql` examples)
* Figure out a way to not nest the statement underneath `EXPLAIN` (e.g. see `12-explains.d/01-base.psql` example)
* Improve handling of nested UNIONs and reduce unnecessary parenthesis (see `35-setops.psql` example)
* Add option to emit long typenames (matching pg_dump)
* (Maybe not even needed) Option to emit pg_catalog schema qualification for built-in data types, functions? (is this important from a security perspective, or does the parser's special handling take precedence?)


postgres_deparse.h
```diff
@@ -0,0 +1,34 @@
+#ifndef POSTGRES_DEPARSE_H
+#define POSTGRES_DEPARSE_H
+
+#include <stdbool.h>
+#include <sys/types.h>
+
+typedef struct PostgresDeparseComment {
+    int after_location;          // Insert comment after the first node with this location
```
lfittl: Per review, could call this `at_location` and clarify that this is emitted (before the node) the first time a node is seen that has at least a location of this value (or more).

src/pg_query_deparse.c
```diff
@@ -21,7 +34,7 @@ PgQueryDeparseResult pg_query_deparse_protobuf(PgQueryProtobuf parse_tree)
 		initStringInfo(&str);
 
 		foreach(lc, stmts) {
-			deparseRawStmt(&str, castNode(RawStmt, lfirst(lc)));
+			deparseRawStmtOpts(&str, castNode(RawStmt, lfirst(lc)), opts);
```
lfittl, line 37: There is a bug here currently where if you had a multi-statement query it would emit comments multiple times (since the comments emitted state is kept local to each `RawStmt`).

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+	{
+		DeparseStateSubPart *subpart = (DeparseStateSubPart *) lfirst(lc);
+		bool first_subpart = list_cell_number(state->result, lc) == 0;
+		bool last_subpart = list_cell_number(state->result, lc) == list_length(state->result) - 1;
+
+		if (state->opts.pretty_print && !first_subpart && (!last_subpart || subpart->str->len > 0))
+		{
+			appendStringInfoChar(str, '\n');
+		}
+		else if (!first_subpart && (state->opts.pretty_print || str->data[str->len - 1] != '('))
```
lfittl: Maybe we can rework this to split pretty print and non-pretty print cases more clearly.

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+deparseEmit(DeparseState *state, StringInfo str)
+{
+	ListCell *lc;
+	foreach (lc, state->result)
+	{
+		DeparseStateSubPart *subpart = (DeparseStateSubPart *) lfirst(lc);
+		bool first_subpart = list_cell_number(state->result, lc) == 0;
+		bool last_subpart = list_cell_number(state->result, lc) == list_length(state->result) - 1;
+
+		if (state->opts.pretty_print && !first_subpart && (!last_subpart || subpart->str->len > 0))
```
lfittl: Could we move the addition of this newline/space to happen after the contents of this subpart were emitted? (to simplify the logic)

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+} DeparseStatePart;
+
+typedef struct DeparseStateStatement
+{
+	List *parts; // List of DeparseStatePart items
 
+	/* Add this much indentation to every subpart */
+	int base_indent;
+
+	struct DeparseStateStatement *parent;
```
lfittl: Can we use a local variable to save the result of `deparseStatePushStatement` (the parent) and then pass that to `deparseStatePopStatement` as an argument? (since it all happens within a function)

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
 	return true;
 }
```
lfittl: Add an explanation here of how we use subparts and parts, and specifically how we flatten parts+subparts into just subparts when moving up in deparseStatePopStatement.

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
-static void deparseAConst(StringInfo str, A_Const *a_const);
-static void deparseGroupingFunc(StringInfo str, GroupingFunc *grouping_func);
-
-static void deparsePreparableStmt(StringInfo str, Node *node);
-static void deparseRuleActionStmt(StringInfo str, Node *node);
-static void deparseExplainableStmt(StringInfo str, Node *node);
-static void deparseStmt(StringInfo str, Node *node);
-static void deparseValue(StringInfo str, union ValUnion *value, DeparseNodeContext context);
+// Each subpart is typically one line to be emitted in pretty print mode
+typedef struct DeparseStateSubPart
```
lfittl: What if we renamed this to `DeparseStatePart` and called the keyword-associated part list `DeparseStateGroup`? (or `DeparseStatePartGroup`)

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+	}
+
+	// Step through all lines, as if we did the final emit, but instead emit line-by-line into the parent
+	foreach (lc, stmt->parts)
+	{
+		DeparseStatePart *part = (DeparseStatePart *) lfirst(lc);
+		bool first_part = list_cell_number(stmt->parts, lc) == 0;
+
+		/* Only merge subparts if they all have the same indentation level */
+		/* TODO: We can be a bit more permissive here - basically we want to avoid merging across statement levels */
```
lfittl: To support this better (e.g. for the `complex_gitlab.sql` test case), we could also alternatively consider keeping a boolean override on the subpart that indicates that the part shall not be merged and must stay on its own line (vs piggybacking on the indent here).

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+			DeparseStateSubPart *subpart = (DeparseStateSubPart *) lfirst(lc2);
+			if (subpart->indent != first_indent)
+			{
+				merge_subparts = false;
+				break;
+			}
+		}
+
+		/* Merge subparts */
+		if (state->opts.pretty_print && merge_subparts && list_length(part->subparts) > 1)
```
lfittl: Should we drop the pretty print flag here? (it would help keep the number of parts down for complex statements, even when not pretty printing)

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+			DeparseStateSubPart *subpart = (DeparseStateSubPart *) lfirst(lc2);
+			*result = lappend(*result, subpart);
+		}
+		list_free(part->subparts);
+		pfree(part);
+	}
+	list_free(stmt->parts);
+
+	// Make sure parent statement writes that follow are on their own line
+	if (stmt->parent && state->opts.pretty_print)
```
lfittl: Can we drop this pretty print conditional too?

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+		deparseAppendStringInfoString(state, ", ");
+	}
+	else
+	{
+		deparseAppendStringInfoChar(state, ',');
+		deparseAppendBreakpoint(state);
+	}
+}
+
+static void
```
lfittl, line 456: Add a comment that talks about what we mean with "major" keyword (since its not a term that is defined by Postgres / SQL Standard).

lfittl: Also, it would be good to rename this to stay more abstract, e.g. `deparseAppendPartGroup` (if we renamed part to part group).

src/postgres_deparse.c
```diff
@@ -130,173 +94,583 @@ isOp(const char *val)
...
+deparseAppendSubpart(DeparseState *state)
+{
+	DeparseStatePart *part = deparseGetCurrentPart(state);
+
+	part->subparts = lappend(part->subparts, makeDeparseStateSubPart(state, state->current, part->indent_subparts));
+}
+
+static void
+deparseAppendBreakpoint(DeparseState *state)
```
lfittl: Rename this to avoid introducing a new term (maybe a `skip_duplicates` argument to deparseAppendSubpart instead?)

postgres_deparse.h
lfittl: I went with `match_location` to help make the connection to the `location` field, instead of someone accidentally thinking of the location/position in the output string.

test/sql/deparse-depesz/08-selects.d/21-in.psql
```diff
@@ -0,0 +1,6 @@
+--- TODO: Should we split the IN list items here?
```
msakrejda, line 1: Something like
```sql
SELECT a
FROM b
WHERE
    c IN (1, 2, 3)
    AND d IN (
        'something longer', 'something longer 2', 'something longer 3',
        'something longer 4', 'something longer 5', ''
    )
```
? Yeah, that might be nice.

test/sql/deparse-depesz/08-selects.d/33-window-functions.psql
```diff
@@ -0,0 +1,16 @@
+--- TODO: Avoid newlines when the contained expressions are simple (single or no keyword)
```
msakrejda: This would be nice to have since `OVER ()` is fairly common.

test/sql/deparse-depesz/08-selects.d/35-setops.psql
```diff
@@ -0,0 +1,15 @@
+--- TODO: The parenthesis around the first UNION are likely unnecessary
```
msakrejda, line 1: +1

test/sql/deparse-depesz/09-inserts.d/03-many-columns.psql
```diff
@@ -0,0 +1,3 @@
+--- TODO: Limit how many items we put on one line
```
msakrejda, line 1: +1

postgres_deparse.h
```diff
@@ -0,0 +1,34 @@
...
+    char *str;                   // The actual comment string, including comment start/end tokens, and newline characters in comment (if any)
+} PostgresDeparseComment;
+
+typedef struct PostgresDeparseOpts {
+    PostgresDeparseComment **comments;
+    size_t comment_count;
+
+    // Pretty print options
+    bool pretty_print;
+    int indent_size;           // Indentation size (Default 4 spaces)
```
msakrejda, line 20: I still think 2 spaces would generally work better for this kind of formatting, but since it's configurable, it's not a big deal.

msakrejda (approved):
I did not look at the code, but I reviewed the examples, and they all look pretty good. I agree with the TODOs, but I think `OVER()` is the only one I'd really miss.

lfittl:
> I did not look at the code, but I reviewed the examples, and they all look pretty good. I agree with the TODOs, but I think `OVER()` is the only one I'd really miss.

Thank you for the review!

test/sql/deparse-depesz/08-selects.d/33-window-functions.psql
lfittl: Agree, resolved this for now by special casing empty window definitions

test/sql/deparse-depesz/08-selects.d/21-in.psql
lfittl: Yup. I will however defer this for now, it seems, based on doing some more investigation just now.

Figuring out how to make this work is non-trivial, and I'd like to get a first version of this code committed soon (and I don't think fixing this will break the API).

test/sql/deparse-depesz/09-inserts.d/03-many-columns.psql
lfittl: FWIW, this is the same issue as the IN list case, so I'll fix those at the same time in a follow-up.

test/sql/deparse-depesz/08-selects.d/35-setops.psql
lfittl: I may or may not address this before merging, but for now I've put this on the "could do in a follow-up" list.

postgres_deparse.h
lfittl: I do find the 4 spaces helpful for tests since it makes the indentation levels clearly visible, but otherwise agree that in practical use 2 spaces is likely more convenient (and that's how we can use this in the app).

src/postgres_deparse.c
```diff
@@ -2477,6 +2848,11 @@ static void deparseSelectStmt(DeparseState *state, SelectStmt *stmt)
 	}
 
 	removeTrailingSpace(state);
+
+	if (need_parens || (context != DEPARSE_NODE_CONTEXT_SELECT_SETOP && context != DEPARSE_NODE_CONTEXT_INSERT_SELECT))
+		deparseStateDecreaseNestingLevel(state, parent_level);
```
msepga, line 2921: Is there a reason the condition here is more complex than `if (parent_level)`?

msepga (approved):
I'm curious about one condition (left a comment), but otherwise LGTM 👍

src/postgres_deparse.c
lfittl: Yeah, `parent_level` can be NULL - I'll add a comment to clarify.

PR #295, opened 2025-07-10
Update GitHub Actions Linux OS to 24.04 for 15-latest

msakrejda:
To support this, merge some of the ci.yml changes from 17-latest:

 - GH actions: Update to actions/checkout@v4
 - Update GH actions cache module to v4
 - Update GitHub Actions Linux OS to 24.04 (#289)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         valgrind: [valgrind,no-valgrind]
     steps:
     - name: Check out code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
     - name: Cache protobuf library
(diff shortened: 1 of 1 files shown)


lfittl:
>  Bump actions workflow to use protobuf 25.1

We may need to skip this one to avoid having to regenerate the protobuf definitions.

msakrejda:
Ah, looks like the Makefile would also need to be udpated to use the new protobuf version. Should I go ahead and do that?

msakrejda:
Trying without the protobuf bump.

msakrejda:
Looks like there's a missing Makefile.msvc--I'll try omitting ea7fac3 as well.

lfittl (approved):

PR #298, opened 2025-08-21
Bump to PostgreSQL 17.6

stanhu:
Update the parser with the latest changes.


lfittl:
> As discussed in https://www.postgresql.org/docs/release/17.6/, PostgreSQL added a `\restrict` command in the output of `pg_dump` and backported this to numerous versions (https://www.postgresql.org/support/security/CVE-2025-8714/). As a result, any dumps generated with these recent versions of PostgreSQL will fail with a parse error.

This doesn't seem right to me, since the regular Postgres parser won't handle psql commands like `\restrict` (and you can see it in the diff, gram.y hasn't actually had any changes short of valgrind-related cleanup).

Not opposed to updating this of course, but can you confirm this actually makes a difference for your use case?

stanhu:
Yeah, I was just trying to confirm that.

You're right; I don't think this update changes that.

Rails is just stripping these lines out: https://github.com/rails/rails/issues/55509 and https://github.com/rails/rails/issues/55510. I suppose that's what needed, and this shouldn't be handled here.

lfittl:
Superseded by #315

PR #300, opened 2025-09-25
[WIP] Add truncation support to `pg_query_summary()`

duckinator:
TODO:
- [x] import `list_sort()`
- [x] import `pg_mbstrlen()`
- [x] import `pg_mbcharcliplen()` and `cliplen()`


duckinator:
Tests _should_ be failing on `test/summary_truncate`, right now.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,409 @@
...
+	rv->alias = NULL;
+	rv->location = 0;
+
+	UpdateStmt *stmt = makeNode(UpdateStmt);
+	stmt->relation = rv;
+	stmt->fromClause = NULL;
+	stmt->targetList = targetList;
+	stmt->whereClause = NULL;
+	stmt->returningList = NULL;
+	stmt->withClause = NULL;
```
lfittl: It appears that `makeNode` internally calls `palloc0` ([Postgres source](https://github.com/postgres/postgres/blob/master/src/include/nodes/nodes.h#L155)), so I don't think we need to initialize anything where the zero value is acceptable (i.e. all these ` = NULL;` lines can be omitted).

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,409 @@
...
+	printf("\n\n! pg_query_summary_truncate()\n");
+	summary_truncation_options(node, &state);
+	return NULL;
+}
+
+static char *truncate_str(char *str, size_t max_chars)
+{
+	// FIXME: This WILL blow up for multi-byte UTF-8 characters, since it's going
+	// by byte instead of character. I think C has functions to deal with this,
+	// I just haven't learned about them yet. -duckinator
```
lfittl: It looks like we can use `pg_mbcliplen` for this, which lets you specify the number of bytes you want to limit to, but respects multi-byte boundaries: https://github.com/postgres/postgres/blob/master/src/backend/utils/mb/mbutils.c#L1084

Alternatively, I wonder if we should treat the truncation limit as number of multi-byte characters, not number of bytes. In that case, `pg_mbcharcliplen` could be used: https://github.com/postgres/postgres/blob/master/src/backend/utils/mb/mbutils.c#L1126 (and we should make sure to use `pg_mbstrlen` to calculate the length of strings)

duckinator: It's now truncating based on the number of characters, since that appears to be what the original intent was.

The functions that are temporarily copied and need to be imported have been added to the TODO list for this PR.

src/pg_query_deparse.c
```diff
@@ -10,6 +10,81 @@
...
+}
+
+PgQueryDeparseResult pg_query_deparse_list(List *stmts)
+{
+	PostgresDeparseOpts opts;
+	MemSet(&opts, 0, sizeof(PostgresDeparseOpts));
+	return pg_query_deparse_opts(stmts, opts);
+}
+
+PgQueryDeparseResult pg_query_deparse_opts(List *stmts, PostgresDeparseOpts opts)
```
lfittl: Thanks for adding this!

I think we could consider refactoring `pg_query_deparse_protobuf_opts` to just call this (after doing pg_query_protobuf_to_nodes), but we can do that in a follow-up PR as well.

src/pg_query_deparse.c
```diff
@@ -10,6 +10,81 @@
...
+		raw->stmt = node;
+		raw->stmt_location = -1;
+		raw->stmt_len = 0;
+		node = (Node *) raw;
+	}
+
+	return pg_query_deparse_list(list_make1(node));
+}
+
+PgQueryDeparseResult pg_query_deparse_list(List *stmts)
```
lfittl: I'd rename this as `pg_query_deparse_stmt_list` to make it clear this must be passed a list of statements, not something else (e.g. a list of expressions is not valid).

src/pg_query_deparse.c
```diff
@@ -10,6 +10,81 @@
 
 #include "protobuf/pg_query.pb-c.h"
 
+PgQueryDeparseResult pg_query_deparse(Node *node)
```
lfittl: I think it'd be better to leave the decision whether its a node (that potentially needs to be wrapped) or a list of RawStmt nodes to the caller, so if we introduce support for deparsing expressions (instead of raw statements) in the future, we don't have a generic function name making the API confusing to understand.

src/pg_query_deparse.c
```diff
@@ -10,6 +10,81 @@
...
+{
+	if (IsA(node, List))
+		return pg_query_deparse_list((List*) node);
+	else
+		return pg_query_deparse_node(node);
+}
+
+PgQueryDeparseResult pg_query_deparse_node(Node *node)
```
lfittl: As discussed, maybe naming this as `pg_query_deparse_stmt` is better here, since I think the node must still be a valid statement (to be wrapped in a `RawStmt`), as compared to e.g. an A_Expr node.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+		return;
+
+	add_truncation(state,
+			TRUNCATION_WHERE_CLAUSE,
+			node,
+			where_clause_len(whereClause));
+}
+
+static bool
+summary_truncation_options(Node *node, TruncationState *state)
```
lfittl: Maybe name this `generate_possible_truncations`?

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+{
+	if (node == NULL)
+		return false;
+
+	switch (nodeTag(node))
+	{
+		case T_RawStmt:
+			// For some reason our statement has been trapped in a RawStmt.
+			// Free them.
```
lfittl: I'd probably skip that comment :-)

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+
+		case T_InsertStmt:
+			{
+				InsertStmt *stmt = castNode(InsertStmt, node);
+
+				if (stmt->cols != NULL)
+					add_truncation(state, TRUNCATION_COLS, node, cols_len(stmt->cols));
+
+				break;
+			}
```
lfittl, line 201: We may want to add support `T_MergeStmt` here as well (which didn't exist when the truncation code was first written, so this is an existing bug, but should be easy to address here)

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+		case T_CopyStmt:
+			{
+				CopyStmt *stmt = castNode(CopyStmt, node);
+
+				add_truncation_where_clause(state, node, stmt->whereClause);
+
+				break;
+			}
+
+		case T_InsertStmt:
```
lfittl: This is an existing issue, but maybe (nitpicking my past self here I guess?) the InsertStmt fits better before the UpdateStmt, so the order is Insert/Update/Merge/Delete/Copy.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+				add_truncation_where_clause(state, node, stmt->whereClause);
+				break;
+			}
+
+		case T_CommonTableExpr:
+			{
+				CommonTableExpr *stmt = castNode(CommonTableExpr, node);
+
+				if (stmt->ctequery != NULL) {
+					PgQueryDeparseResult result = pg_query_deparse(node);
```
lfittl: I think this needs to be `(Node *) stmt->ctequery`, so we're actually passing something that is a statement.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+
+					if (result.error) {
+						state->error = result.error;
+						return false;
+					}
+
+					int32_t length = strlen(result.query);
```
lfittl: I'd probably inline this in the function call below, to match the other `_len` calls (though this is a different kind of function being called, I don't think putting it in a variable is necessary)

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+{
+	fprintf(stderr, "ERROR from %s:\n  %s:%s:%i: %s\n",
+			func, error->funcname, error->filename, error->lineno, error->message);
+}
+
+static int32_t select_target_list_len(List *nodes) {
+	PgQueryDeparseResult result = pg_query_deparse_node(dummy_select(nodes, NULL, NULL));
+
+	if (result.error) {
+		print_pg_query_error(result.error, __func__);
```
lfittl: We could investigate if we can make `elog(ERROR, ..)` work here, so the whole truncation fails, instead of making it not truncate that portion. Because we work with a known input (a valid parse tree), we can require that the deparser always succeeds, and consider it a bug if it doesn't.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+		default:
+			break;
+	}
+
+	if (!pg_query_raw_tree_walker_supports(node))
+		return false;
+
+	state->depth++;
+
+	return raw_expression_tree_walker(node, summary_truncation_options, (void *) state);
```
lfittl: As discussed, this won't correctly behave when there is sibling nodes that the walker gets called for, so we'd either need to pass this to the function as a separate argument (seems challenging to implement), or restore the previous value after having descended down, like this:

```suggestion
	int old_depth = state->depth;
	state->depth++;

	bool result = raw_expression_tree_walker(node, summary_truncation_options, (void *) state);

	// Restore old depth value since the current node (or its parents) may have sibling elements
	state->depth = old_depth;

	return result;
```

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+	return pt_a->depth - pt_b->depth;
+}
+
+static void expand_ellipses(char *str)
+{
+	// https://www.compart.com/en/unicode/U+2026
+	// pattern is a double quote, ellipses, and another double quote.
+	char pattern[5] = {'"', 0xE2, 0x80, 0xA6, '"'};
+
+	for (size_t i = 0; i < strlen(str); i++) {
```
lfittl: I think we'd need to subtract the pattern length from strlen here, so we don't overrun when the pattern would no longer fit into the remaining string.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+
+	if (!pg_query_raw_tree_walker_supports(node))
+		return false;
+
+	state->depth++;
+
+	return raw_expression_tree_walker(node, summary_truncation_options, (void *) state);
+}
+
+static int cmp_possible_truncation_depth(const ListCell *a, const ListCell *b)
```
lfittl: As discussed, the intent here is to "Truncate the deepest possible truncation that is the longest first" (see [Ruby implementation](https://github.com/pganalyze/pg_query/blob/main/lib/pg_query/truncate.rb#L16)), so we're missing the length of the truncation in this comparison. I think we can probably just rename this function `cmp_possible_truncation_depth` and add length to it (I think you'd just check depth first, return early if its different, and if its the same then compare length, making sure that longer things sort first, i.e. the opposite logic we have for depth).

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+	PgQueryDeparseResult result = pg_query_deparse(tree);
+	char *output = result.query;
+
+	if (result.error)
+		return result.error;
+
+	if (strlen(result.query) <= truncation_limit) {
+		summary->truncated_query = output;
+		return NULL;
+	}
```
lfittl: We could do this before generating truncations to save on the generation if the query is already short enough.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+			RuleStmt *stmt = castNode(RuleStmt, node);
+			stmt->whereClause = (Node *) dummy_column();
+			printf("Rule/whereClause\n");
+		}
+		else if (IsA(node, CommonTableExpr) && attr == TRUNCATION_CTE_QUERY) {
+			// Equivalent to: https://github.com/pganalyze/pg_query.rs/blob/66eb7becea1a40e315fee3f90197a35a89d20c25/src/truncate.rs#L230-L237
+			CommonTableExpr *stmt = castNode(CommonTableExpr, node);
+			Node *old = stmt->ctequery;
+			stmt->ctequery = dummy_select(NULL, (Node *) dummy_column(), NULL);
+			if (old) {
```
lfittl: To capture my understanding: This logic (that came from Rust originally, and is not present in the Ruby implementation) helps us not revisit the statement contained (e.g. `SelectStmt`), because we already removed the whole CTE anyway. Because it is valid to update a removed node (since we do not free it, we just remove it from the tree), this should not affect correctness, but rather it lets us potentially skip an unnecessary deparse.

However, I think because we go depth first, this may never be utilized, so its potentially worth dropping here.

lfittl: Also, if we remove this, we could replace the iteration logic with a foreach, since we don't actually have to modify the list as we iterate over it.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+	while (list_length(state->truncations) > 0) {
+		// Get the first item from the list.
+		PossibleTruncation *truncation = linitial(state->truncations);
+		// ... and then remove it from the list.
+		state->truncations = list_delete_nth_cell(state->truncations, 0);
+
+		Node *node = truncation->node;
+		enum TruncationAttr attr = truncation->attr;
+
+		if (truncation->length <= 3) {
```
lfittl: It might be worth avoiding creating these truncations in the first place.

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+			// If "truncating" would make it longer, refuse to truncate.
+		} else if (IsA(node, SelectStmt) && attr == TRUNCATION_TARGET_LIST) {
+			SelectStmt *stmt = castNode(SelectStmt, node);
+			stmt->targetList = list_make1(dummy_target());
+			printf("Select/targetList\n");
```
lfittl: Per our discussion, these were just for debugging and will go away :)

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,673 @@
...
+		else if (IsA(node, InsertStmt) && attr == TRUNCATION_COLS) {
+			InsertStmt *stmt = castNode(InsertStmt, node);
+			stmt->cols = list_make1(dummy_target());
+			printf("Insert/cols\n");
+		}
+		else if (IsA(node, IndexStmt) && attr == TRUNCATION_WHERE_CLAUSE) {
+			IndexStmt *stmt = castNode(IndexStmt, node);
+			stmt->whereClause = (Node *) dummy_column();
+			printf("Index/whereClause\n");
+		}
```
lfittl: If we add support for MergeStmt, we should of course also add it here.

src/postgres_deparse.c
```diff
@@ -11869,6 +11872,9 @@ static void deparseStmt(DeparseState *state, Node *node)
 		case T_CommentStmt:
 			deparseCommentStmt(state, castNode(CommentStmt, node));
 			break;
+		case T_CommonTableExpr:
```
lfittl: This should not be needed once we fix the case where we pass `node` instead of `stmt->ctequery` to deparse (when generating truncations).

src/postgres_deparse.c
```diff
@@ -2730,6 +2730,9 @@ static void deparseUtilityOptionList(DeparseState *state, List *options)
 
 static void deparseSelectStmt(DeparseState *state, SelectStmt *stmt, DeparseNodeContext context)
 {
+	if (stmt == NULL)
```
lfittl: Its not clear to me why that would happen, might be worth removing and seeing what needs this (if anything).

test/framework/main.c
```diff
@@ -164,18 +178,39 @@ TEST_ASSERT_LIST_EQUAL_impl(TestState * test_state, char *actual_str, List *actu
 	free(exp_ary);
 }
 
+// Given a list of tests, cleanup function, and use_mctx,
+// run tests and return the results.
+//
+// If `use_mctx` is true, a Postgres MemoryContex is enter before the test is run,
```
lfittl:
```suggestion
// If `use_mctx` is true, a Postgres MemoryContex is entered before the test is run,
```

src/pg_query_summary_truncate.c
duckinator: I looked into this and as best I can tell it's guaranteed to never overrun just because of how `memcmp` compares things.

The pattern is `"…"`, and it's 5 bytes. The edge case we're worried about is it being at the end of the string.

These are the only outcomes I could manage to create:

1. it makes 0 or more successful matches, then a null byte, and `memcmp(str + i, pattern, 5) == 0` returns a non-zero result immediately after encountering the null byte.
2. it has a syntax error, and never gets to `expand_ellipses()`.

lfittl: Good point, I didn't think about the null byte comparison. Agreed this should be safe as is :)

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,684 @@
...
+	// https://www.compart.com/en/unicode/U+2026
+	// pattern is a double quote, ellipses, and another double quote.
+	char pattern[5] = {'"', 0xE2, 0x80, 0xA6, '"'};
+
+	for (size_t i = 0; i < strlen(str); i++) {
+		if (memcmp(str + i, pattern, 5) == 0) {
+			size_t len = strlen(str + i + 5);
+			memcpy(str + i, "...", 3);
+			memmove(str + i + 3, str + i + 5, len + 1);
+		}
```
lfittl:
```suggestion
size_t pattern_len = 5;
char* replace = "...";
size_t replace_len = strlen(replace);
...
		if (memcmp(str + i, pattern, pattern_len) == 0) {
		    char* remainder = str + i + pattern_len;
			memcpy(str + i, replace, replace_len);
			memmove(str + i + replace_len, remainder, strlen(remainder) + 1);
		}
```

src/pg_query_summary_truncate.c
```diff
@@ -0,0 +1,684 @@
...
+	char *pattern = "SELECT ... AS ...";
+	size_t plen = strlen(pattern);
+
+	size_t slen = strlen("SELECT ...");
+
+	for (size_t i = 0; i < strlen(str); i++) {
+		if (memcmp(str + i, pattern, plen) == 0) {
+			size_t len = strlen(str + i + plen);
+			memmove(str + i + slen, str + i + plen, len + 1);
+		}
```
lfittl:
```suggestion
		if (memcmp(str + i, pattern, plen) == 0) {
		    char* remainder = str + i + plen;
			memmove(str + i + slen, remainder, strlen(remainder) + 1);
		}
```

src/pg_query_summary.c
```diff
@@ -619,11 +621,10 @@ pg_query_summary_walk(Summary * summary, Node *tree)
...
 	 *
-	 * - We process cte_names as we iterate through the tree.
-	 * - We only add items to tables if they are *not* in cte_names.
-	 * - CTEs can be defined *after* they're referenced as names.
-	 * - Because of this, we have to store range_vars and process it after
-	 *   the initial tree walk.
+	 * - We process cte_names as we iterate through the tree. - We only add
+	 * items to tables if they are *not* in cte_names. - CTEs can be defined
+	 * *after* they're referenced as names. - Because of this, we have to
+	 * store range_vars and process it after the initial tree walk.
```
duckinator: This was messed up by the reformatting.

src/pg_query_summary.c
```diff
@@ -789,14 +775,14 @@ summary_to_protobuf(PgQuerySummaryParseResult * result, Summary * summary)
...
-	 * }
-	 */
+	if (summary->truncated_query == NULL)
+	{
+		sr.truncated_query = pstrdup("");
+	}
+	else
+	{
+		sr.truncated_query = summary->truncated_query;
+	}
```
duckinator: Maybe this instead?

```suggestion
	if (summary->truncated_query == NULL)
		sr.truncated_query = pstrdup("");
	else
		sr.truncated_query = summary->truncated_query;
```

test/summary_truncate.c
```diff
@@ -0,0 +1,215 @@
...
+#[should_panic(
+    expected = "byte index 22 is not a char boundary; it is inside 'は' (bytes 21..24) of `WITH \"原チコ氏にはす腹腹腹腹腹腹腹腹腹腹腹\" AS (SELECT) SELECT w`"
+)]
+*/
+/*
+void byte_truncate_fails(TestState* test_state) {
+    char* query = "WITH \"原チコ氏にはす腹腹腹腹腹腹腹腹腹腹腹\" AS (SELECT) SELECT w".to_string();
+    query[0..=21].to_string();
+}
+*/
```
duckinator: I'm not sure how to re-add this. I don't think the C test infrastructure supports an equivalent right now.

duckinator: Actually, I'm not sure this test does anything useful. It seems like it's testing how Rust works, not implementation details?

src/pg_query_summary.c
duckinator:
```suggestion
	 * NOTE regarding cte_names and range_vars:
	 * - We process cte_names as we iterate through the tree.
	 * - We only add items to tables if they are *not* in cte_names.
	 * - CTEs can be defined *after* they're referenced as names.
	 * - Due to this, we store range_vars and process it after the tree walk.
```

test/summary_truncate.c
seanlinsley: Yes it was added at the same time as the test above it, documenting string truncation behavior in Rust: https://github.com/pganalyze/pg_query.rs/pull/7

duckinator: Okay, so since we're truncating to a certain number of characters (not bytes) we don't have that problem. Thanks for confirming that. ^^

lfittl (approved):
:+1:

PR #310, opened 2025-11-21
Add test framework support for running individual tests.

duckinator:
For test files using `test/framework/` (currently only `test/summary.c` and `test/summary_truncate.c`), you can now run an individual test by specifying it as the first argument.

E.g.:

```console
puppy@cerberus:~/dev/libpg_query$ ./test/summary it_parses_INSERT
it_parses_INSERT (line 834)

test result: ok. 17 passed; 0 failed; 0 skipped
puppy@cerberus:~/dev/libpg_query$ 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,9 +63,6 @@ ifeq ($(VALGRIND),1)
 	override CFLAGS += -DUSE_VALGRIND
 	override TEST_CFLAGS += -DUSE_VALGRIND
 endif
-ifeq ($(TEST_FAIL_FAST),1)
-	override TEST_CFLAGS += -DTEST_FAIL_FAST=1
-endif
 
 CLEANLIBS = $(ARLIB)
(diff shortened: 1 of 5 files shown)


duckinator:
I think this is fine as-is, but it might be a good opportunity to replace the `TEST_FAIL_FAST` variable with a `-ff` flag or similar. This PR already weaves argc + argv through to `test_run_impl()`, so most of the changes are already done.

lfittl:
> I think this is fine as-is, but it might be a good opportunity to replace the `TEST_FAIL_FAST` variable with a `-ff` flag or similar. This PR already weaves argc + argv through to `test_run_impl()`, so most of the changes are already done.

Yeah, seems reasonable. I think we can add a special case for `-ff` here without introducing full opt parsing, since this is just a test command.

duckinator:
This is good to go. :+1:

lfittl (approved):

PR #311, opened 2025-11-21
Add summary support for MERGE statements 

duckinator:


duckinator:
It's worth noting that I did modify the test slightly. The original test expected the statement types to to be just `{"MergeStmt"}`, but the code was returning `{"MergeStmt", "SelectStmt"}` which seems _more correct_ to me for this query:

```
WITH cte AS (SELECT * FROM g.other_table CROSS JOIN p) MERGE INTO my_table USING cte ON (id=oid) WHEN MATCHED THEN UPDATE SET a=b WHEN NOT MATCHED THEN INSERT (id, a) VALUES (oid, b);
```

lfittl (approved):

PR #312, opened 2025-11-22
Fix compilation warnings for the examples in README.md

neshkeev:
I noticed a couple of warnings when compiling the examples from `README.md`
Please consider merging my changes or tell me how to make the PR better


lfittl (approved):
:+1: Good catch, thanks for the improvements!

PR #313, opened 2025-11-29
Add `pg_query_is_utility_stmt()`

duckinator:
Add a C API that can be used by https://github.com/pganalyze/collector/blob/main/util/is_utility.go

### DIFF
--- a/pg_query.h
+++ b/pg_query.h
@@ -19,6 +19,7 @@ typedef struct {
 typedef struct {
 	int length;
 	bool *items;
+	PgQueryError* error;
 } PgQueryIsUtilityResult;
 
 typedef struct {
@@ -140,6 +141,8 @@ PgQuerySplitResult pg_query_split_with_parser(const char *input);
 PgQueryDeparseResult pg_query_deparse_protobuf(PgQueryProtobuf parse_tree);
 PgQueryDeparseResult pg_query_deparse_protobuf_opts(PgQueryProtobuf parse_tree, struct PostgresDeparseOpts opts);
 PgQueryDeparseCommentsResult pg_query_deparse_comments_for_query(const char *query);
+
+PgQueryIsUtilityResult pg_query_is_utility_stmt(const char *query);
 
 PgQuerySummaryParseResult pg_query_summary(const char* input, int parser_options, int truncate_limit);
 
--- a/pg_query.h
+++ b/pg_query.h
@@ -1,6 +1,7 @@
 #ifndef PG_QUERY_H
 #define PG_QUERY_H
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <sys/types.h>
 
@@ -14,6 +15,11 @@ typedef struct {
 	int cursorpos; // char in query at which exception occurred
 	char* context; // additional context (optional, can be NULL)
 } PgQueryError;
+
+typedef struct {
+	int length;
+	bool *items;
+} PgQueryIsUtilityResult;
 
 typedef struct {
   size_t len;
@@ -137,6 +143,7 @@ PgQueryDeparseCommentsResult pg_query_deparse_comments_for_query(const char *que
 
 PgQuerySummaryParseResult pg_query_summary(const char* input, int parser_options, int truncate_limit);
 
+void pg_query_free_is_utility_result(PgQueryIsUtilityResult result);
 void pg_query_free_normalize_result(PgQueryNormalizeResult result);
 void pg_query_free_scan_result(PgQueryScanResult result);
 void pg_query_free_parse_result(PgQueryParseResult result);
--- a/pg_query.h
+++ b/pg_query.h
@@ -146,7 +146,6 @@ PgQueryIsUtilityResult pg_query_is_utility_stmt(const char *query);
 
 PgQuerySummaryParseResult pg_query_summary(const char* input, int parser_options, int truncate_limit);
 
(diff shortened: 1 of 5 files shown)


src/pg_query_is_utility_stmt.c
```diff
@@ -0,0 +1,48 @@
...
+pg_query_is_utility_stmt_walk(Node *node, List **result)
+{
+	if (node == NULL)
+		return false;
+
+	switch (nodeTag(node)) {
+		case T_SelectStmt:
+		case T_InsertStmt:
+		case T_UpdateStmt:
+		case T_DeleteStmt:
```
lfittl, line 19: This is missing MergeStmt (which to be fair is also missing in the collector code, but we should fix that here)

src/pg_query_is_utility_stmt.c
```diff
@@ -0,0 +1,200 @@
...
+#include <nodes/nodes.h>
+#include <nodes/nodeFuncs.h>
+
+typedef struct {
+	List *result;
+} UtilityWalkState;
+
+static
+bool
+pg_query_is_utility_stmt_walk(Node *node, UtilityWalkState *state)
```
lfittl:
```suggestion
pg_query_is_utility_stmt_walk_actual(RawStmt *raw_stmt)
```

+ don't descend into child nodes (only check the top-level `raw_stmt->stmt` type)

src/pg_query_is_utility_stmt.c
```diff
@@ -0,0 +1,200 @@
...
+PgQueryIsUtilityResult
+pg_query_is_utility_stmt(const char *query)
+{
+	PgQueryIsUtilityResult result = {0};
+	MemoryContext ctx = pg_query_enter_memory_context();
+
+	PgQueryInternalParsetreeAndError parsetree_and_error = pg_query_raw_parse(query, 0);
+
+	UtilityWalkState state = {NULL};
+	pg_query_is_utility_stmt_walk((Node *) parsetree_and_error.tree, &state);
```
lfittl:
```suggestion
    foreach (lc, parsetree_and_error.tree) {
      RawStmt *raw_stmt = lfirst_node(lc, RawStmt);
      state->result = lappend(state->result, pg_query_is_utility_stmt_actual(raw_stmt));
    }
```

src/pg_query_is_utility_stmt.c
```diff
@@ -0,0 +1,58 @@
...
+	size_t length = list_length(parsetree_and_error.tree);
+	result.length = 0;
+	result.items = malloc(sizeof(bool) * length);
+
+	ListCell *lc;
+	foreach(lc, parsetree_and_error.tree) {
+		RawStmt *raw_stmt = lfirst_node(RawStmt, lc);
+		result.items[result.length] = is_utility_stmt_actual(raw_stmt);
+		result.length++;
+	}
```
lfittl:
```suggestion
	result.length = list_length(parsetree_and_error.tree);
	result.items = malloc(sizeof(bool) * result.length);

	ListCell *lc;
	foreach(lc, parsetree_and_error.tree) {
		RawStmt *raw_stmt = lfirst_node(RawStmt, lc);
		result.items[foreach_current_index(lc)] = is_utility_stmt_actual(raw_stmt);
	}
```

Feels slightly better to me since it avoids the separate `length` variable.

src/pg_query_is_utility_stmt.c
```diff
@@ -0,0 +1,58 @@
...
+}
+
+PgQueryIsUtilityResult
+pg_query_is_utility_stmt(const char *query)
+{
+	PgQueryIsUtilityResult result = {0};
+	MemoryContext ctx = pg_query_enter_memory_context();
+
+	PgQueryInternalParsetreeAndError parsetree_and_error = pg_query_raw_parse(query, 0);
```
lfittl, line 34: I think we have to check the `error` in `parsetree_and_error` here, if the statement that was passed has a syntax error.

i.e. do something like this:

```c
if (parsetree_and_error.error)
{
    result.error = parsetree_and_error.error;
}
else
{
   // Existing Code below that does the work
} 

if (parsetree_and_error.stderr_buffer)
    free(parsetree_and_error.stderr_buffer);

pg_query_exit_memory_context(ctx);

return result;
```

pg_query.h
```diff
@@ -137,6 +143,7 @@ PgQueryDeparseCommentsResult pg_query_deparse_comments_for_query(const char *que
 
 PgQuerySummaryParseResult pg_query_summary(const char* input, int parser_options, int truncate_limit);
 
+void pg_query_free_is_utility_result(PgQueryIsUtilityResult result);
```
lfittl: Shouldn't we also add the regular function to this file somewhere?

i.e. what you currently have in `is_utility_stmt.c`:

```c
extern PgQueryIsUtilityResult pg_query_is_utility_stmt(const char *query);
```

test/is_utility_stmt.c
```diff
@@ -0,0 +1,127 @@
+#include <pg_query.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "test/framework/main.h"
+
+extern PgQueryIsUtilityResult pg_query_is_utility_stmt(const char *query);
```
lfittl: Once we add this to `pg_query.h` (see separate comment) it should not be necessary to declare it here.

lfittl (approved):

PR #321, opened 2026-01-02
C API for parsing, deparsing and fingerprinting queries

levkk:
Bypasses Protobuf (de)ser and allows bindings to access the `libpg_query` API via the C ABI. Used in a fork of `pg_query.rs` to accelerate `parse` and `deparse`.

More context: https://github.com/pganalyze/pg_query.rs/issues/72

Note: there is a bit of copy/paste going on here because I wanted to leave the protobuf paths untouched. Let me know if you'd prefer I merge them to re-use more code.

### DIFF
--- a/pg_query.h
+++ b/pg_query.h
@@ -135,6 +135,26 @@ void pg_query_exit(void);
 #define PG_VERSION "16.1"
 #define PG_VERSION_NUM 160001
 
+// Raw parse tree access (bypasses protobuf serialization)
+// Note: The returned tree uses PostgreSQL's memory context. The tree is only
+// valid until pg_query_free_raw_parse_result is called or pg_query_exit is called.
+
+// Forward declaration of PostgreSQL List type (defined in nodes/pg_list.h)
+struct List;
+// Forward declaration of PostgreSQL MemoryContextData type
+struct MemoryContextData;
+
+typedef struct {
+  struct List *tree;  // PostgreSQL parse tree (List of RawStmt nodes)
+  char* stderr_buffer;
+  PgQueryError* error;
+  struct MemoryContextData* context;  // Internal: Memory context for the tree (do not modify)
+} PgQueryRawParseResult;
+
+PgQueryRawParseResult pg_query_parse_raw(const char* input);
+PgQueryRawParseResult pg_query_parse_raw_opts(const char* input, int parser_options);
+void pg_query_free_raw_parse_result(PgQueryRawParseResult result);
+
 // Deprecated APIs below
 
 void pg_query_init(void); // Deprecated as of 9.5-1.4.1, this is now run automatically as needed
--- a/pg_query.h
+++ b/pg_query.h
@@ -201,6 +201,29 @@ void *pg_query_list_make1(void *datum);
 void *pg_query_list_append(void *list, void *datum);
 PgQueryDeparseResult pg_query_deparse_nodes(void *stmts);
 
+// Raw scan (bypasses protobuf serialization)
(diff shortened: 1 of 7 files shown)


src/pg_query_deparse.c
```diff
@@ -85,6 +85,188 @@ pg_query_free_deparse_result(PgQueryDeparseResult result)
...
+		foreach(lc, stmts)
+		{
+			deparseRawStmtOpts(&str, castNode(RawStmt, lfirst(lc)), opts);
+			if (lnext(stmts, lc))
+				appendStringInfoString(&str, "; ");
+		}
+		result.query = strdup(str.data);
+	}
+	PG_CATCH();
+	{
```
NikolayS, line 171: shouldn't we have `MemoryContextSwitchTo` here?

src/pg_query_scan.c
```diff
@@ -172,3 +172,137 @@ void pg_query_free_scan_result(PgQueryScanResult result)
...
+    {
+      if (core_yylex(&yylval, &yylloc, yyscanner) == 0) break;
+    }
+    scanner_finish(yyscanner);
+
+    // Allocate output array
+    result.tokens = malloc(sizeof(PgQueryRawScanToken) * token_count);
+    result.n_tokens = token_count;
+
+    // Second pass: fill in token data
```
NikolayS, line 231: couldn't it be done with just a single pass? (using a growable array - eg, start with 64, then realloc 2x when needed)

levkk: This is just a copy/paste of the existing implementation, sans protobuf conversion.

lfittl: Yeah, the existing implementation here is lazy, since its cheap enough to just run twice. We could totally modify this to use realloc, but I don't think we need to do it in this PR.

src/pg_query_deparse.c
```diff
@@ -85,6 +85,188 @@ pg_query_free_deparse_result(PgQueryDeparseResult result)
...
+	return (void *) pg_query_enter_memory_context();
+}
+
+void
+pg_query_deparse_exit_context(void *ctx)
+{
+	pg_query_exit_memory_context((MemoryContext) ctx);
+}
+
+/* Node allocation helper */
```
lfittl, line 108: Hmmm, I see why you added this and the other methods, but I do wonder if this is the right approach (e.g. why not access the include header that has `palloc0` directly?)

src/pg_query_deparse.c
lfittl: Yeah, I agree, that seems like an oversight - CopyErrorData [will fail](https://github.com/pganalyze/libpg_query/blob/17-latest/src/postgres/src_backend_utils_error_elog.c#L1516) when the context isn't switched.

pg_query.h
```diff
@@ -166,6 +166,64 @@ void pg_query_exit(void);
 #define PG_VERSION "17.7"
 #define PG_VERSION_NUM 170007
 
+// Raw parse tree access (bypasses protobuf serialization)
```
lfittl, line 169: I think it'd be better if we put all of this into `pg_query_raw.h`, so authors of bindings that do not use the direct access to the AST structs don't get confused which methods to use.

lfittl (commented):
> Note: there is a bit of copy/paste going on here because I wanted to leave the protobuf paths untouched. Let me know if you'd prefer I merge them to re-use more code.

Yeah, the copy/paste isn't great, it would be good to reduce duplication. I think we should be good to call a shared "raw" implementation in many of these cases.

PR #323, opened 2026-01-22
feat: add plpgsql proto support

psteinroe:
> This pull request was written alongside Claude Code. As a fellow os maintainer, I can completely understand if it is closed without comments.

Adds support for parsing plpgsql into protobuf, replacing the experimental json output. I do not have much experience in C or Ruby but tried to follow existing patterns as close as possible. I verified it by pointing the Rust binding in the Postgres Language Server project at my fork. Tests pass, and it seems to work just fine (https://github.com/supabase-community/postgres-language-server/pull/639).

This would enable much better plpgsql support in the Language Server! :) 

Summary of changes:
- Add pg_query_parse_plpgsql_protobuf() API function
- Generate PLpgSQL types in pg_query.proto
- Add automated code generation scripts for PLpgSQL nodes
- Include outfuncs/readfuncs for serialization


lfittl:
@psteinroe Thanks for the contribution, and appreciate the proactive disclosure of having used Claude Code - I think that's not a deal breaker as long as you have tested/reviewed the result.

I've been looking at this already as part of the Postgres 18 PR - I will take your work here into account (and may use parts of it), and will try to take this into the Postgres 18 update that I'm actively working on completing.

lfittl:
@psteinroe I ended up running out of time to get this into the initial 18 release, but I think this still makes sense to do.

In case you have time available to work on this in this PR after all, two things that'd be helpful:

1) Rebase on Postgres 18 (against 18-latest branch)
2) Extend this to also auto-generate the JSON output functions, so we can address #333 as well

PR #324, opened 2026-01-23
pg_query_deparse_comments_for_query: Add missing initialization

lfittl:
This caused result.error to be undefined when there was no error (instead of NULL), which can cause a crash in the caller when reading it.

### DIFF
--- a/src/pg_query_deparse.c
+++ b/src/pg_query_deparse.c
@@ -88,7 +88,7 @@ pg_query_free_deparse_result(PgQueryDeparseResult result)
 PgQueryDeparseCommentsResult
 pg_query_deparse_comments_for_query(const char *query)
 {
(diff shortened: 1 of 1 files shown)


keiko713:
Do you have some example of this fails/goes well?

lfittl:
> Do you have some example of this fails/goes well?

We don't really exercise this in our current tests in libpg_query, but in the Ruby library you can easily confirm the fix by running `bundle exec rspec ./spec/lib/pg_query/deparse_pretty_print_spec.rb:5` on main, and then changing [the same file](https://github.com/pganalyze/libpg_query/blob/17-latest/src/pg_query_deparse.c#L91) and then rebuilding and re-running the test.

keiko713 (approved):

PR #328, opened 2026-02-03
Makefile: Update `AR` command to include `ARFLAGS`

MisterDA:
Otherwise it's difficult to correctly set `AR` when invoking make. This is the default rule: `$(AR) $(ARFLAGS) $@ $<`. The `ARFLAGS` variable is documented: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-ARFLAGS. Its default is `rv`.


lfittl (approved):
:+1: Makes sense, thanks for the contribution!

Issue #334, opened 2026-05-19
Normalize NOTIFY commands


seanlinsley:
Currently, `NOTIFY` commands generate a unique query fingerprint for every unique payload passed as the second argument. That should be fixed, so that the payload value is no longer included in the fingerprint.
```rb
PgQuery.fingerprint("NOTIFY channel, 'a'")
=> "f2410f11c52cd1d9"
PgQuery.fingerprint("NOTIFY channel, 'b'")
=> "2a8ec734fe3c6490"
```

We may also want to generate a unique fingerprint based on the channel name (currently that's not included in the fingerprint). Though we'd want to try and [normalize the channel name](https://github.com/pganalyze/libpg_query/commit/ce2959ff19dd462e18423428c1bd052b67b2545c).

Additionally, normalizing a `NOTIFY` command currently does nothing. It should be updated to remove the payload value.
```rb
PgQuery.normalize("NOTIFY channel, 'data that should be removed'")
=> "NOTIFY channel, 'data that should be removed'"

duckinator:
Actionable tasks:
1. for fingerprinting a NOTIFY command, remove the payload entirely
2. for normalizing a NOTIFY command, replace the payload with (e.g. '$1')

As per discussion with @seanlinsley, I'll be skipping normalizing random channel names for now.

----

Good starting point would be adding some unit tests.

There's an existing fingerprint test for `NOTIFY virtual` — I could just add a few with data passed, and make sure they all return the same value: https://github.com/pganalyze/libpg_query/blob/204fbdbd3ed5f8691ab358e49f1fc5397b4679e2/test/fingerprint_tests.c#L136-L137


There is no existing normalize (or normalize_utility) test for `NOTIFY`, but it should be fairly easy to add:

https://github.com/pganalyze/libpg_query/blob/204fbdbd3ed5f8691ab358e49f1fc5397b4679e2/test/normalize_tests.c
https://github.com/pganalyze/libpg_query/blob/204fbdbd3ed5f8691ab358e49f1fc5397b4679e2/test/normalize_utility_tests.c

PR #340, opened 2026-05-28
Normalize NOTIFY commands

duckinator:
Addresses #334

### DIFF
--- a/src/pg_query_normalize.c
+++ b/src/pg_query_normalize.c
@@ -567,6 +567,23 @@ static bool const_record_walker(Node *node, pgssConstLocations *jstate)
 				if (jstate->normalize_utility_only) return false;
 				return raw_expression_tree_walker(node, const_record_walker, (void*) jstate);
 			}
+		case T_NotifyStmt:
+			{
+				NotifyStmt *stmt = castNode(NotifyStmt, node);
+
+				if (stmt->payload == NULL)
+					break; // No payload to normalize.
+
+				char *loc = strstr(jstate->query, ",");
+
+				if (loc == NULL)
+					// Somehow there's a payload but no comma?
+					// This should be impossible.
+					break;
+
+				record_defelem_arg_location(jstate, loc - jstate->query + 1);
+				break;
+			}
 		case T_InsertStmt:
 			{
 				if (jstate->normalize_utility_only) return false;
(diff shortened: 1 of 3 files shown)


duckinator:
@seanlinsley this is ready for review. :+1:

src/pg_query_normalize.c
```diff
@@ -567,6 +567,23 @@ static bool const_record_walker(Node *node, pgssConstLocations *jstate)
...
+
+				if (stmt->payload == NULL)
+					break; // No payload to normalize.
+
+				char *loc = strstr(jstate->query, ",");
+
+				if (loc == NULL)
+					// Somehow there's a payload but no comma?
+					// This should be impossible.
+					break;
```
seanlinsley, line 582: What's the story with the guard clauses on line 574 and 579? Were these added so the tests would pass?

test/normalize_tests.c
```diff
@@ -42,6 +42,14 @@ const char* tests[] = {
   "MERGE into measurement m USING new_measurement nm ON (m.city_id = nm.city_id and m.logdate=nm.logdate) WHEN MATCHED AND nm.peaktemp IS NULL THEN DELETE WHEN MATCHED THEN UPDATE SET peaktemp = greatest(m.peaktemp, nm.peaktemp), unitsales = m.unitsales + coalesce(nm.unitsales, $1) WHEN NOT MATCHED THEN INSERT (city_id, logdate, peaktemp, unitsales) VALUES (city_id, logdate, peaktemp, unitsales)",
   "DO language E'plpgsql' 'BEGIN PERFORM 1; END'",
   "DO language $1 $2",
+  "NOTIFY channel, 'a'",
+  "NOTIFY channel, $1",
```
seanlinsley, line 46: It's worth noting that this normalization turns it into invalid SQL, causing a parse to fail with:
```
syntax error at or near "$1" (scan.l:1262)
```

@lfittl do we expect that normalized queries are parseable, and should the parser be updated to accept this? Or should we store SQL that's currently valid during normalization?
```sql
-- Store an empty string
NOTIFY channel, ''
-- Or remove the payload entirely
NOTIFY channel

src/pg_query_normalize.c
duckinator: The one on 574 is because for `NOTIFY` statements the payload is optional: https://www.postgresql.org/docs/current/sql-notify.html

The only way I can think of that being represented is the payload being NULL, since it being an empty string would be something different. If there's no payload, there's no normalization to happen here.

The one on 579 is just my usual semi-defensive programming style — if I call a function that can return `NULL`, and I don't control the input, I tend to assume that a `NULL` might be returned at some point.

test/normalize_tests.c
lfittl: I think keeping it as a parameter reference is best, and we can fix our parser to support that.

duckinator: @seanlinsley if we're keeping this as a parameter reference, I believe that means this PR is done (but requires parser changes), right?

(I'm going to do the fingerprinting as a separate PR.)

seanlinsley: Yes that's right. Please also update the parser so it accepts this new syntax.

seanlinsley (approved):

PR #348, opened 2026-06-21
Add stack overflow crash protection, error out instead

lfittl:
This makes libpg_query return the standard "stack depth limit exceeded" error that Postgres also shows for overly deep queries, instead of crashing when stack depth gets exceeded. This error is triggered by appropriately placed calls to the Postgres check_stack_depth() function, and goes together with initializing the stack base and depth in pg_query_init.

The maximum stack depth is handled similar to Postgres max_stack_depth variable, defaulting to 100 kB, and auto-sized up to 2MB. Because we may be called in threads with varying stack sizes, and from different host program locations, the allowed stack depth is recalculated for each call to libpg_query.

Per multiple reports from OSS-Fuzz.

## Follow-ups

* `protobuf_c_message_unpack` can run into similar issues, but since its a vendored library its not exactly sure how we should deal with it (possibly have a set of patches on top of it, and automate the vendoring?)
* The PG_TRY/PG_CATCH sections in the parser are repetitive, especially in how they populate the result - can we unify them, or add a function?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,7 +238,7 @@ examples/normalize_error: examples/normalize_error.c $(ARLIB)
 examples/simple_plpgsql: examples/simple_plpgsql.c $(ARLIB)
 	$(CC) $(TEST_CFLAGS) -o $@ -g examples/simple_plpgsql.c $(ARLIB) $(TEST_LDFLAGS)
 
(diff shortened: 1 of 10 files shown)


seanlinsley:
Looks like there's a memory leak

lfittl:
> Looks like there's a memory leak

Yep, that was in Protobuf C++ specifically. Fixed!

PR #344, opened 2026-06-21
Avoid needing testLength for fingerprint tests.

duckinator:
This was being calculated wrong, and thus not all tests were running.

Unfortunately, there are now failing tests.

`tests/fingerprint_tests.c` is now indented wrong, but I felt it made sense to get the tests working before re-indenting that, to keep the diff as small as possible.

PR #350, opened 2026-06-25
Remove need for testsLength for everything except fingerprint tests.

duckinator:


duckinator:
I left the fingerprint tests alone in this PR; there's some open questions I mentioned over in #344.

I wound up taking a different approach for this PR, because it allows all of the individual files to be valid code while still removing the need for `testsLength`.

test/fingerprint_opts_tests.c
```diff
@@ -29,6 +29,7 @@ const char* tests[] = {
   "NEW.author.first_name = upper(cleanString(NEW.author.first_name))",
   "5",
   "a148e3f78b53c252",
-};
 
-size_t testsLength = __LINE__ - 4;
+  // Trailing NULL to tell the test runner we're done.
+  NULL
```
seanlinsley: Could we put the NULL and the comment on the same line? (in all files)

duckinator: Done. :+1:

seanlinsley (approved):

PR #353, opened 2026-07-14
Don't include NOTIFY payload in fingerprints.

duckinator:
`_fingerprintNotifyStmt` previously fingerprinted the payload, which meant NOTIFY statements with different payloads produced different fingerprints.
    
Now the payload is ignored, bringing it in line with existing (intentional) behavior for the channel name of focusing on the structure, not data.


src/include/pg_query_fingerprint_defs.c
```diff
@@ -13667,13 +13667,8 @@ _fingerprintRuleStmt(FingerprintContext *ctx, const RuleStmt *node, const void *
 static void
 _fingerprintNotifyStmt(FingerprintContext *ctx, const NotifyStmt *node, const void *parent, const char *field_name, unsigned int depth)
```
lfittl, line 13668: I think we need to update the script that generates it instead, roughly here: https://github.com/pganalyze/libpg_query/blob/18-latest/scripts/generate_fingerprint_outfuncs.rb#L315 and then re-run the script.

duckinator: Done. :+1:

seanlinsley (commented):
Could you add tests to confirm the fingerprint stays the same with different payloads?

duckinator:
@seanlinsley sorry about that, looks like they got lost when I re-did the PR. added the tests I had originally. :+1:

seanlinsley (approved):

PR #356, opened 2026-07-21
PG17: Serialize retvarno and out_param_varno in PL/pgSQL parse output

pyramation:
I built and tested this against the full 23,000 statements in pgsql-parser, and a slew of other tests... 

The JSON output of `pg_query_parse_plpgsql` omits two fields needed to faithfully reconstruct RETURN / RETURN NEXT statements:

- `PLpgSQL_stmt_return.retvarno` / `PLpgSQL_stmt_return_next.retvarno` — without it, `RETURN NEXT <variable>` is indistinguishable from a bare `RETURN NEXT;`, so any consumer round-tripping the AST back to source drops the returned variable.
- `PLpgSQL_function.out_param_varno` — needed to tell the two apart in OUT-parameter functions, where the compiler sets `retvarno` implicitly even for a bare `RETURN NEXT;` (and where emitting the variable name back is invalid syntax).

This adds both fields to the serialized output. Existing output is unchanged except for the added keys.


lfittl:
@pyramation Thanks for the contribution!

Can you remind me why you opened this against 17? Is that because of the current challenges with PL/pgSQL parsing on the 18 branch, and this way you're unblocked on having this working?

pyramation:
I was doing it to keep things in sync, but we can close this and focus on 18 from here on out. I'm fine with 18 :)

Issue #335, opened 2026-05-19
Normalize random table names during fingerprinting


seanlinsley:
[Some query workloads](https://github.com/cedar-team/django-bulk-load) generate queries with random table names. We should identify those and generate a single query fingerprint for the same query structure, even if the table name has a random substring.

For example:
- `loading_books_dc786da1fead11f` has `dc786da1fead11f` which should be normalized
- `_temp__2577316832_vk95nrgo_users` has both `2577316832` and `vk95nrgo` which should be normalized

This can be done by checking the randomness of each underscore-delimited substring in the table name, and skipping those characters when fingerprinting the table name.

Note that there is some prior work on this subject, which was intended to normalize tables that use timeseries partitioning: https://github.com/pganalyze/libpg_query/commit/ce2959ff19dd462e18423428c1bd052b67b2545c

I wonder if we should simply avoid fingerprinting any substring which is more than two characters long, and contains a number.

duckinator:
@seanlinsley would it be possible to get a collection of real-world examples (even if you change the non-random parts) I can use as a starting point for writing more tests for this?

seanlinsley:
It's a bit difficult to find real-world examples because it involves scanning a lot of data. Let's try to implement something that works for the examples in the issue description for now, if that's okay with you.

duckinator:
Alright. 👍

PR #361, opened 2026-08-21
Fingerprinting: Introduce options to preserve Postgres 17 fingerprint compatibility

lfittl:
This intends to assist with situations where pg_query processes data from a Postgres 17 or older database, that has the prior queryid handling of considering schema and table names significant, but ignoring aliases. The pg_query fingerprint is aimed to be a superset of the Postgres queryid, and this allows callers to upgrade libpg_query but continue matching fingerprinting behavior to the Postgres release actually in use. This assumption broke in the libpg_query 18 release due to us following the [upstream change](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=787514b30bb7dd0b3484d6cb717e3b9aafc06c4a) for queryids.

Additionally, allow controlling the fingerprinting of relation names to turn off the current behavior of ignoring 2+ digits in table names, which is also aimed to support upcoming work in #335 for handling random table names better.

This is an API and ABI breaking change for the `pg_query_fingerprint_opts` function, since it gains an additional argument - we should bump the minor release version accordingly.


msepga:
Looked this over and tested things locally. LGTM, and the new option makes sense 👍

msepga (approved):

seanlinsley:
I think users could benefit from a persistent markdown file (or a section in the README) explaining the rationale for our fingerprint defaults, and in what cases they should consider changing the defaults. Or at least a detailed CHANGELOG entry, because I'm not sure this PR is enough to go on.

PR #366, opened 2026-09-13
Guard libpg_query's own recursive walkers against stack exhaustion

hellower:
Deeply nested but otherwise valid SQL makes libpg_query recurse until the OS thread stack runs out, and the process **dies with SIGSEGV instead of returning an error**. On `18-latest` today:

```c
/* SELECT 1+1+1+... with 64000 terms, 128KB of input */
PgQueryParseResult r = pg_query_parse(sql);   /* never returns: SIGSEGV */
```

That matters for any caller that parses untrusted SQL, and it is not recoverable from a managed runtime — in Go the signal arrives during a cgo call, so `recover()` never sees it.

I ran into this in a Postgres-compatible server that feeds client SQL to `pg_query_go`, where a single statement could take the daemon down. I carried the fix in a fork first, then ported it here.

## Why it happens — two things combine

**1. The PostgreSQL stack-depth guard is shipped but disabled.**

`check_stack_depth()`, `stack_is_too_deep()` and `max_stack_depth` are all present in `src/postgres/src_backend_utils_misc_stack_depth.c`. But `stack_is_too_deep()` only reports true when `stack_base_ptr` is non-NULL, and the body of `set_stack_base()` — the function that sets it — has been stripped, leaving the declaration in `miscadmin.h` and an empty `#ifdef` shell in the `.c` file:

```c
/*
 * set_stack_base: set up reference point for stack depth checking
 ...
 */
#ifndef HAVE__BUILTIN_FRAME_ADDRESS
#endif
#ifdef HAVE__BUILTIN_FRAME_ADDRESS
#else
#endif
```

`assign_max_stack_depth()` was stripped the same way, so `max_stack_depth_bytes` is stuck at the 100kB compile-time default.

**2. libpg_query's own recursive walkers never call the guard.**

Counting `check_stack_depth()` call sites on `18-latest`:

| file | origin | calls |
|---|---|---|
| `src_backend_nodes_copyfuncs.c` | from PostgreSQL | 1 |
| `src_backend_nodes_nodeFuncs.c` | from PostgreSQL | 1 |
| `src_backend_nodes_equalfuncs.c` | from PostgreSQL | 1 |
| **`pg_query_outfuncs_protobuf.c`** | **libpg_query's own** | **0** |
| **`pg_query_outfuncs_json.c`** | **libpg_query's own** | **0** |
| **`pg_query_readfuncs_protobuf.c`** | **libpg_query's own** | **0** |
| **`postgres_deparse.c`** | **libpg_query's own** | **0** |
| **`vendor/protobuf-c/protobuf-c.c`** | **vendored** | **0** |

So even with the guard armed, nothing on the paths that actually overflow would consult it. Fixing (1) alone changes essentially nothing — I measured that before finding (2).

## What this PR does

- **Restores** `set_stack_base()`, `restore_stack_base()` and `assign_max_stack_depth()` bodies, verbatim from PostgreSQL.
- **Arms the guard in `pg_query_enter_memory_context()`**, which every entry point already goes through — no entry point needs to change.
  The limit is **derived from the running thread's stack** rather than assumed. PostgreSQL can rely on the main stack plus a GUC; a library linked into an arbitrary host cannot. macOS uses `pthread_get_stacksize_np()`; Linux (glibc and musl) uses `RLIMIT_STACK` on the main thread and `pthread_getattr_np()` + `pthread_attr_getstack()` on others; Windows uses `GetCurrentThreadStackLimits()`. Half the stack, clamped to `[64kB, 1GB]`, with a 2MB fallback elsewhere (e.g. the BSDs). **If you would rather expose this as an explicit API** (`pg_query_set_max_stack_depth()`), the derivation here can become the default instead of the only behaviour — happy to change it.
- **Calls `check_stack_depth()`** in `_outNode` (both serializers), `_readNode`, `deparseExpr`/`deparseStmt` and normalize's `const_record_walker` — the single dispatcher each walker passes through per nesting level, so one call covers the whole tree. (`const_record_walker` also had a catch-all `PG_CATCH` that swallowed the new error and returned a partially normalized query as success; it now re-throws it.)
- **protobuf-c**: the unpack path uses roughly 1kB of C stack per level, so a count-only limit is not enough; it gets both `check_stack_depth()` and a nesting counter (`PROTOBUF_C_MAX_UNPACK_NESTING`, 10000). The counter mirrors what protobuf-go and protobuf-c++ already do — upstream protobuf-c has no recursion limit at all. The check goes in `get_packed_size` rather than `pack` because the former runs *before* the malloc, so a longjmp there leaks nothing.
  🚨 This touches vendored third-party code, so it is the part most likely to want a different home — tell me if you would rather carry it as a patch, or take it to protobuf-c upstream.
- **Serialization gets its own `PG_TRY`.** `pg_query_raw_parse()` has one, but it ends before serialization starts, so an `ereport()` raised while walking the tree had no exception stack to unwind to and PostgreSQL escalated it to **FATAL** — the process exited instead of returning an error. This became reachable the moment the walkers started calling `check_stack_depth()`.
- **Unpack failure** raised an `Assert`; it now raises `ERRCODE_STATEMENT_TOO_COMPLEX` so release builds return an error too, and `free_unpacked()` is passed the same allocator the unpack used.
- **`USE_PROTOBUF_CPP` builds** get the same guarantee without longjmp'ing across C++ frames: the C++ walker records that the tree is too deep and the error is raised after the protobuf messages are destroyed. It also caps nesting at one level per kB of `max_stack_depth`, because the protobuf C++ runtime's own recursion (serialize, JSON, destructors) uses far more stack per level than the walker.
- **`pg_query_summary()`** no longer leaks the parser's stderr buffer when a walk raises.
- **`test/stack_depth`** (new, wired into `Makefile` and `Makefile.msvc`) runs every tree-walking entry point on input far deeper than a 1MB stack allows, on a thread created with a 1MB stack, and requires `stack depth limit exceeded`; shallow and wide inputs must still succeed. Removing the Windows stack-size branch makes it fail on all six Windows jobs, at `test/stack_depth` (MSVC: `0xc00000fd`, stack overflow; MSYS2: exit 127).

## Measured (darwin/arm64, subprocess-isolated)

`SELECT 1+1+…` is the cheapest shape per byte. Same binary otherwise:

| input | before | after |
|---|---|---|
| 6KB (3000 terms) | parsed | parsed |
| 48KB (24000 terms) | parsed | parsed |
| 128KB (64000 terms) | **SIGSEGV** | `stack depth limit exceeded` |
| 500KB (250000 terms) | **SIGSEGV** | `stack depth limit exceeded` |

**Large but shallow input keeps working**, which is the point of bounding depth rather than length: `SELECT true OR true …` ×100000 (800KB, nesting depth 11) and `SELECT ((((1))))` ×8000 both still parse normally.

`make test` passes, including `test/stack_depth`: on darwin/arm64, on linux/amd64 with GCC 14 (manylinux_2_28), on Alpine (musl), and across this repository's full CI matrix run on my fork (clang/gcc × protobuf-c/protobuf-cpp × valgrind, macOS, MSVC x64/x86, MSYS2 ×4): https://github.com/hellower/libpg_query/actions/runs/34877094171 (same tree as this branch)

## Notes

- Issue #9 (2016) has a similar title but is a different bug — `_outNode` recursing forever on `CreateForeignTableStmt` because of a struct embedding. This is unbounded depth on input that is deep but otherwise valid.
- There is no `SECURITY.md` or private reporting channel on the repo, so I am reporting this together with the fix rather than separately.
- Reproducer used for the table above:

```c
#include <pg_query.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static char *chain(int n) {
	char *s = malloc(9 + (size_t) n * 2);
	strcpy(s, "SELECT 1");
	for (int i = 0; i < n; i++) strcat(s, "+1");
	return s;
}

int main(int argc, char **argv) {
	int n = argc > 1 ? atoi(argv[1]) : 64000;
	char *sql = chain(n);
	PgQueryParseResult r = pg_query_parse(sql);
	printf("%s\n", r.error ? r.error->message : "parsed ok");
	pg_query_free_parse_result(r);
	free(sql);
	return 0;
}
```


lfittl:
@hellower Thank you for the report - I believe this is the same as an issue previously detected by OSS-fuzz. Despite the SIGSEGV crash we don't believe this is actively exploitable.

The planned fix is to switch to the upb library instead of protobuf-c, see #349 - I intend to land this in the next 1-2 weeks. If you want to take a look at that PR and validate that this addresses the concern that would be appreciated.

Also thank you for calling out the missing clarity on how to report potentially security sensitive issues - I'll see that we clarify that. For now in case you find anything else, feel free to report privately to security@pganalyze.com

hellower:
@lfittl Thanks for the pointer, and for clarifying the reporting channel. I tested #349 (head `8f3319b`) against `18-latest` (`7632d03`) and this branch.

**Short version: #349 bounds the protobuf decode, but most of the crash is not on that path, so it doesn't address this.** The JSON output, normalize and summary walkers are unchanged in #349 and overflow exactly as before. Protobuf encoding still overflows too: in #349 the frame at the crash is upb's `encode_field`, whose depth limit is set to `0xFFFF`, far deeper than a thread stack can hold.

Setup: darwin/arm64, `make build` for each tree, the same harness linked against each `libpg_query.a`, every case in its own process, the call made on a pthread with an explicit stack size. macOS reports a thread stack overflow as SIGBUS.

| entry point | input | stack | `18-latest` | #349 | this PR |
|---|---|---|---|---|---|
| `pg_query_parse` (JSON) | `SELECT 1+1…` ×64000 | 8 MB | SIGBUS | **SIGBUS** | `stack depth limit exceeded` |
| `pg_query_parse_protobuf` | ×24000 | 8 MB | SIGBUS | **SIGBUS** | `stack depth limit exceeded` |
| `pg_query_parse_protobuf` | ×3000 | 512 kB | SIGBUS | **SIGBUS** | `stack depth limit exceeded` |
| `pg_query_normalize` | ×24000 | 8 MB | SIGBUS | **SIGBUS** | `stack depth limit exceeded` |
| `pg_query_summary` | ×250000 | 8 MB | SIGBUS | **SIGBUS** | `stack depth limit exceeded` |
| `pg_query_deparse_protobuf` | ×10000 | 8 MB | SIGBUS | `could not parse protobuf` | `stack depth limit exceeded` |

The last row is the one #349 fixes. Everything above it crashes identically on `18-latest` and #349.

While testing #349 I found three issues in it that are independent of this PR:

- its decode depth limit rejects fairly ordinary queries (a 23-column `||` concatenation, 16 levels of scalar subquery);
- `pg_query_parse_protobuf()` can return an empty tree without an error;
- `pg_query_parse_protobuf()` overflows at a shallower depth than on `18-latest` (18,746 vs 23,861 terms on an 8 MB stack).

Details and numbers: https://github.com/pganalyze/libpg_query/pull/349#issuecomment-5667112465

**The two PRs look complementary rather than overlapping.** Once #349 lands, the protobuf-c part of this PR becomes moot: the vendored `protobuf-c.c`/`.h` changes, the allocator plumbing, and the unpack error handling. I'm happy to rebase onto #349 and drop all of that, leaving the stack base setup and the `check_stack_depth()` calls in libpg_query's own walkers. (The encoder would still need a limit that fits the stack; the `_outNode` guard bounds the tree depth, but upb's encoder then recurses over that tree on the same stack, so I'd rather measure whether that is enough after rebasing than assume it.) The rebase would also remove the part of this PR most likely to need discussion: the edits to vendored code.

Testing also turned up a gap in this PR, now fixed:

- **`552dfe8`** — normalize. `const_record_walker()` recurses into itself directly for `SelectStmt`, so a long `UNION ALL` chain never reached a stack check (still SIGBUS at 5000 branches on a 512 kB stack). Worse, its default case wraps the walk in a `PG_TRY` that flushes every error, so once the guard existed it swallowed `stack depth limit exceeded` too: `pg_query_normalize()` reported success with only part of the constants replaced. That same `PG_TRY` also returned from inside the try block, leaving `PG_exception_stack` pointing at a dead frame. All three are fixed, and each change was reverted on its own to confirm it's needed. `make test` passes.
- **`cba0ad0`** — comment text only: removes the fork-specific labels I had left on the comments.

One trade-off I should be upfront about: the limit here is half the thread stack (clamped to `[64kB, 1GB]`), so it is 4 MB on an 8 MB stack, and this PR rejects some inputs `18-latest` handles (e.g. deparsing a 5000-branch `UNION ALL`, which needs more than 4 MB of stack but fits in 8 MB). If you'd rather make that configurable, or derive it differently, I'm happy to change it. The explicit `pg_query_set_max_stack_depth()` option from the description is still on the table.

_Edit: corrected the limit above. It is capped at 1 GB, not 1 MB; the PR description and the first commit message said 1 MB as well. Both are fixed now (the branch was rebuilt to reword that commit, so the two commits above have new hashes); the code was always 1 GB._

hellower:
I've force-pushed this branch. The first CI run on it had failed, and fixing that turned up a few more problems. Summary of what changed (details in the commit messages and the updated description):

- **Build fixes for the CI failures.** clang and GCC 14 rejected `pthread_getattr_np()` without `_GNU_SOURCE`. MSVC couldn't parse `__thread` in `protobuf-c.h`, and my declaration there had also ended up between `PROTOBUF_C__API` and `protobuf_c_message_unpack()`. `pthread.h` was included unconditionally, which MSVC would have hit next.
- **The guard now works on Windows and musl.** Both fell back to a 2MB limit, which can't fire on Windows' 1MB default stack or musl's ~130kB thread stacks. Windows now uses `GetCurrentThreadStackLimits()`. Linux uses `RLIMIT_STACK` on the main thread (musl's `pthread_getattr_np()` only reports the part mapped so far) and `pthread_getattr_np()` on other threads.
- **New `test/stack_depth`.** It found two more problems. The `USE_PROTOBUF_CPP` serializer had no guard; it needs a non-longjmp version plus a depth cap, because the protobuf C++ runtime's recursion is much more stack-hungry than the walker. And `pg_query_summary()` leaked the parser's stderr buffer when a walk raised.
- The first commit is reworded so its message says 1GB, not 1MB.

The full CI matrix passes on my fork against the same tree: https://github.com/hellower/libpg_query/actions/runs/34877094171. With the Windows stack-size branch removed, the new test fails on all six Windows jobs.

The CI run for this push needs your approval to start.

PR #367, opened 2026-09-17
Summary: Fix a relation going missing when a CTE shares its name

aurelienbottazini:
Related PR on Ruby gem: https://github.com/pganalyze/pg_query/pull/355

`pg_query_summary()` drops a relation when a CTE shares its name:

```c
"WITH t AS (SELECT * FROM s.t) SELECT 1"   /* tables: []   expected: s.t */
"WITH t AS (SELECT * FROM t) SELECT 1"     /* tables: []   expected: t   */
```

`WalkState.cte_names` is a single flat set for the whole statement, and
`handle_range_var()` treated every range var whose `relname` matched one of those
names as a CTE reference. Three separate cases fall out of that:

1. A non-recursive CTE is not visible inside its own definition, so in
   `WITH users AS (SELECT * FROM users) SELECT * FROM users` the inner reference
   is the `users` table and only the outer one is the CTE. Both were skipped.
2. Only `relname` was compared, so a schema qualified reference was skipped too
   (the first example above).
3. A CTE cannot be the target of `INSERT`/`UPDATE`/`DELETE`/`MERGE` or DDL, so
   those always name a real relation. `WITH users AS (...) UPDATE users SET ...`
   reported no target table.

This is easy to hit from ORMs that name a CTE after the model's own table, and it
matters for anything using the table list for query attribution, read/write
routing or cache invalidation, because the relation goes missing silently.

### Fix

* `handle_with_clause()` runs before descending into the statement, so the CTE names of enclosing scopes are
  known by the time a nested `WITH` clause is reached.
* It records the range vars inside each CTE's own definition that carry the CTE's
   own name (`cte_self_reference_walker()`), and `handle_range_var()` skips only
  those. The scan is skipped for`WITH RECURSIVE` (which does make a CTE visible to itself) and when the name is
  already a visible CTE from an enclosing scope, e.g. the inner `a` in
  `WITH a AS (...) SELECT * FROM (WITH a AS (SELECT * FROM a) ...) s`.
* `is_cte_reference()` now gathers the whole decision: `CONTEXT_SELECT` only,
  unqualified only, name in `cte_names`, and not a recorded self-reference.

Range vars are still resolved after the tree walk, and the order of
`summary->tables` and `summary->cte_names` is unchanged, so no existing
expectations move.

### Tests

13 cases in `test/summary_tests.c` (`test/summary_tests_list.c` regenerated with
`scripts/update_summary_tests_list.rb`)

Against unpatched `src/`, 9 of them fail (20 assertions). With the fix,
`make test` is green

### Performance

Measured through the Ruby binding, with these files dropped into pg_query's
`ext/pg_query/` and `PgQuery.summary` called 20k times per query:

| query | before | after |
| --- | --- | --- |
| `SELECT * FROM companies WHERE id = 1` | 1.91us | 1.89us |
| `WITH x AS (SELECT * FROM companies) SELECT * FROM x` | 2.08us | 2.14us |
| `WITH companies AS (SELECT * FROM companies) SELECT * FROM companies` | 1.90us | 2.07us |
| 3-CTE query with joins | 7.23us | 7.78us |
| `SELECT * FROM t0` + 100 joins | 36.0us | 37.1us |

Statements without a `WITH` clause are unaffected; statements with one pay a
single extra pass over the CTE definitions.

### Notes / open questions

No `CHANGELOG.md` entry here, matching recent fix PRs on this branch. Glad to
  add one wherever you'd like it.


lfittl:
Thanks! I've asked @duckinator to take a look at this.

duckinator (approved):
This looks good to me. I'm not surprised there were a few bugs lurking in the CTE handling; it had a lot of edge cases.

seanlinsley (approved):

